Issue #122, opened 2019-07-31
File uploader input file types is broken


esmailpour-hosein:
It shows every file type no matter what types are provided!
It doesn't add the file types to `access` attribute!

PR #123, opened 2019-07-31
Fix file uploader input accepted file types

esmailpour-hosein:
accepted file type was unnecessarily required and fixed that! Also for the accepted file types there was an issue with the conditional attribute which is fixed now!
Fixes issue #122

@zdlm

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
     "lerna": "^3.13.4",
     "lint-staged": "^8.1.6",
     "lit-element": "^2.1.0",
+    "lit-html": "^1.1.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -131,6 +131,7 @@
     "lerna": "^3.13.4",
     "lint-staged": "^8.1.6",
     "lit-element": "^2.1.0",
+    "lit-html": "^1.1.1",
```
zdlm: Why do you install lit-html. OK, for `ifDefined `

esmailpour-hosein: Yeah, exactly

PR #129, opened 2019-08-09
Refactor Typography component

esmailpour-hosein:
- Replace color with type to make it consistent with other components
- Add disabled and disabled-active type colors
- Add tooltip support when the text is added by slot

### DIFF
--- a/packages/components/typography/src/utils/constants.js
+++ b/packages/components/typography/src/utils/constants.js
@@ -4,7 +4,9 @@ export const colorTypes = {
 	DANGER: 'danger',
 	ERROR: 'error',
 	SUCCESS: 'success',
-	WARNING: 'warning'
+	WARNING: 'warning',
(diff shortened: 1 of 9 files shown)


packages/components/typography/src/utils/constants.js
```diff
@@ -4,7 +4,9 @@ export const colorTypes = {
 	DANGER: 'danger',
 	ERROR: 'error',
 	SUCCESS: 'success',
-	WARNING: 'warning'
+	WARNING: 'warning',
+	DISABLED: 'disabled',
+	DISABLED_ACTIVE: 'disabled-active'
```
zdlm: What is `disabled_active`? weird type

esmailpour-hosein: For situations like menu, which can be active or checked and still disabled!

PR #147, opened 2019-09-03
Fix typography truncated text issue in display flex parent

esmailpour-hosein:


index.css
```diff
@@ -54,3 +54,8 @@ ts-list-item {
 	display: block;
 	max-width: var(--ts-box-width-medium);
 }
+
+.typography-wrapper {
+	display: flex;
+	max-width: 350px;
```
zdlm: Where is the number come from? If it is common stuff, let's put it into variable!

esmailpour-hosein: No, it's just for test matters inside the index.html.

Issue #155, opened 2019-09-13
Add onOpen on the ts-aside


CarloAl:
We have already defined in `constants.js` the `onOpen` and `onOpened` events, but they are never used in `aside.js` (while the `onClose` events are actually triggered). Can we add the same for `onOpen?`

esmailpour-hosein:
I cleaned up those in this pr https://github.com/Tradeshift/elements/pull/154.
For the `onOpen` event I don't think it make sense to have it since developers know when they are actually adding the `data-visible` attribute. But for the `onOpened `(`opened` after that PR) maybe we can add the event if we have a good use case for it. Did you encountered a situation that you needed that event?

CarloAl:
In my case I have an aside that is used to select the visible columns in a table. Every time I open the aside I need to refresh the updated selected columns and because of the way our code is structured I don't have scope of the columns where I open the aside. I could use custom events but I guess it is pretty standard to have an `onpen` event in this case.

PR #157, opened 2019-09-18
Add onOpen and onOpened in the aside

zdlm:
@esmailpour-hosein 

#155 

```
			function onOpen() {
				console.log('The aside is opening');
			}

			function onOpened() {
				console.log('The aside is opened');
			}

			function onClose() {
				console.log('The aside is closing');
			}

			function onClosed() {
				console.log('The aside is closed');
			}

			document.getElementById('my-aside').addEventListener('onOpen', onOpen);
			document.getElementById('my-aside').addEventListener('onOpened', onOpened);
			document.getElementById('my-aside').addEventListener('onClose', onClose);
			document.getElementById('my-aside').addEventListener('onClosed', onClosed);
```


index.html
```diff
@@ -674,6 +674,14 @@ <h2>Typography</h2>
 			<ts-cover id="my-cover"></ts-cover>
 		</ts-root>
 		<script>
+			function onOpen() {
```
esmailpour-hosein: We need to come up with a better and more consistent solution to show the functionality in future 🤔

zdlm: We can use Open and Close as you refactored before.

packages/components/aside/src/aside.js
```diff
@@ -69,5 +67,29 @@ customElementDefineHelper(
 				<ts-cover class="ts-aside-cover" ?data-visible=${this.visible} @click="${this.close}"></ts-cover>
 			`;
 		}
+
+		update(changedProperties) {
+			super.update(changedProperties);
+			if (changedProperties.has('visible')) {
+				const oldVal = changedProperties.get('visible');
+				if (oldVal === false) {
```
esmailpour-hosein: Can you refactor this to something like this?
const eventName = oldVal ? customEventNames.CLOSE :  customEventNames.OPEN;
this.dispatchCustomEvent(eventName)

zdlm: This is not if else. The first time it is undefined. I can't trigger Close .

Issue #159, opened 2019-09-25
Create new platform object slot in the aside 


CarloAl:
Some of the aside will be used to display platform object like the following:

![image](https://user-images.githubusercontent.com/5701042/65582283-6599e000-df7d-11e9-969f-85bfb6de0c0c.png)

Add a slot with padding and grey background color where platform object can be displayed.

PR #160, opened 2019-09-25
Add platform object slot to aside

CarloAl:


packages/components/aside/src/aside.css
```diff
@@ -46,6 +50,13 @@
 	padding-right: var(--ts-unit);
 }
 
+/* Platform object.................................................. */
+
+.aside-platform-object {
+	padding: calc(var(--ts-unit) / 2);
```
zdlm: var(--ts-unit-half)

packages/components/aside/src/aside.js
```diff
@@ -68,6 +71,13 @@ customElementDefineHelper(
 			`;
 		}
 
+		handleSlotPlatformObjectChange(e){
+			const slot = e.currentTarget;
+			const assignedNodes = slot.assignedNodes();
+			const showSlot = assignedNodes && assignedNodes.length;
+			slot.parentElement.classList[showSlot ? 'remove' : 'add']('hidden');
```
esmailpour-hosein: Can you just set an internal property (same as hasFoot) here and then do the conditional class name in the 
```
<div class="aside-platform-object ${this.hasPlatformObject ? '' : 'hidden'}">
```

CarloAl: Done. Any reason why it is preferred to classList approach (which I took from `root.js`), is it a question of elegance? Also should we not have the same for footer? If the footer is removed (because the same aside element is used to show different aside, one with footer and the other without?) it will still be displayed.

esmailpour-hosein: As you can see in `root.js` file the `decorateSlot` was used in lots of slots, so it tries to don't duplicate the code! 
The `defaultSlotChangeHandler` is a bit different and can be approached in both way but the difference here is you have a default class name in place which is better to just show the logic in there to make it easier to follow

esmailpour-hosein:
Fix the issue  https://github.com/Tradeshift/elements/issues/159

PR #173, opened 2019-10-10
Use spinner in the aside

zdlm:
#170 
Screenshot 2019-10-10 at 13 56 34

### DIFF
--- a/packages/components/aside/src/aside.css
+++ b/packages/components/aside/src/aside.css
@@ -19,6 +19,24 @@
 	display: none;
 }
 
+/* Spinner........................................................ */
+.spinner-overlay {
+	position: absolute;
+	top: var(--ts-unit-triple);
+	right: 0;
+	bottom: 0;
+	left: 0;
+	background-color: rgba(255, 255, 255, 0.7);
+}
+
+.aside-container ts-spinner {
+	position: absolute;
+	top: 50%;
+	left: 50%;
+	margin-top: calc(-0.5 * var(--spinner-size-large));
+	margin-left: calc(-0.5 * var(--spinner-size-large));
+}
+
 /* Header......................................................... */
 
 .aside-container header {
--- a/packages/components/aside/src/aside.js
+++ b/packages/components/aside/src/aside.js
@@ -2,6 +2,7 @@ import { TSElement, unsafeCSS, html, customElementDefineHelper } from '@tradeshi
 import css from './aside.css';
 import '@tradeshift/elements.button';
 import '@tradeshift/elements.cover';
+import '@tradeshift/elements.spinner';
 import { customEventNames } from './utils';
 
 customElementDefineHelper(
@@ -15,6 +16,7 @@ customElementDefineHelper(
 			return {
 				title: { type: String, attribute: 'data-title' },
 				visible: { type: Boolean, attribute: 'data-visible', reflect: true },
+				busy: { type: String, attribute: 'data-busy', reflect: true },
 				hasFoot: { type: Boolean },
 				hasPlatformObject: { type: Boolean }
 			};
@@ -53,6 +55,17 @@ customElementDefineHelper(
 			return `${pre}-${mid}-${dir}`;
 		}
 
+		get spinner() {
+			if (this.busy === undefined || this.busy === null) {
+				return '';
+			}
+			return html`
+				<div class="spinner-overlay">
+					<ts-spinner data-visible data-message="${this.busy}"></ts-spinner>
+				</div>
+			`;
+		}
+
 		render() {
 			return html`
 				<div dir="${this.direction}" class="aside-container ${this.slide} ${this.hasFoot ? 'has-footer' : ''}">
@@ -65,7 +78,7 @@ customElementDefineHelper(
 					<div class="aside-note">
 						<slot name="note" class="note-in-aside"></slot>
 					</div>
-					<div class="aside-platform-object ${this.hasPlatformObject ? '' : 'hidden'}"">
+					<div class="aside-platform-object ${this.hasPlatformObject ? '' : 'hidden'}">
 						<slot name="platform-object" @slotchange="${this.platformObjectSlot}"></slot>
 					</div>
 					<main class="aside-main">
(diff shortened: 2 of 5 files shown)


lloydhumphreys:
Should this dim the content somehow? an opaque white overlay?

zdlm:
Good idea. I will make an opaque white overlay. We can change it if @mihneaz likes diming content

esmailpour-hosein:
> Should this dim the content somehow? an opaque white overlay?

As far as I know we don't have a design for it but I agree as well. @mihneaz what do you say?
Not sure about it being white

zdlm:
Screenshot 2019-10-10 at 14 42 30

packages/components/aside/src/aside.js
```diff
@@ -15,6 +16,7 @@ customElementDefineHelper(
 			return {
 				title: { type: String, attribute: 'data-title' },
 				visible: { type: Boolean, attribute: 'data-visible', reflect: true },
+				busy: { type: String, attribute: 'data-busy', reflect: true },
```
esmailpour-hosein: why it's string? For the reason Jakob had issue with boolean properties?

packages/components/aside/src/aside.js
```diff
@@ -65,7 +76,7 @@ customElementDefineHelper(
 					<div class="aside-note">
 						<slot name="note" class="note-in-aside"></slot>
 					</div>
-					<div class="aside-platform-object ${this.hasPlatformObject ? '' : 'hidden'}"">
```
esmailpour-hosein: hmmm, our linter shouldn't stop this?

packages/components/aside/src/aside.js
```diff
@@ -65,7 +78,7 @@ customElementDefineHelper(
...
```
esmailpour-hosein: hmmm, our linter shouldn't stop this?

mihneaz:
Looks ok for the aside but I checked https://tradeshift.github.io/elements/?path=/story/ts-spinner--default and I noticed few other things we need to address:
- The back color (the color of the complete circle) has to be @ts-color-gray-light with 50% opacity. The opacity ensures it looks ok against both white backgrounds and an overlay.
- The small version needs to be 3px in thickness and the medium needs to be 4px
- I think we need a knob to showcase it with a dark overlay (that'll default the color knob to white)

packages/components/aside/src/aside.css
```diff
@@ -19,6 +19,16 @@
 	display: none;
 }
 
+/* Spinner........................................................ */
+
+.aside-container ts-spinner {
+	position: absolute;
+	top: 50%;
+	left: 50%;
+	margin-top: calc(-0.5 * var(--spinner-size-large));
```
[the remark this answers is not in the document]
zdlm: Yes, It works. we use the vars in calc in our vars.css

packages/components/aside/src/aside.js
```diff
@@ -53,6 +55,15 @@ customElementDefineHelper(
 			return `${pre}-${mid}-${dir}`;
 		}
 
+		get spinner() {
+			if (this.busy === undefined) {
```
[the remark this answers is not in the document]
zdlm: If it doesn't have the property.  I won't add the html

packages/components/aside/src/aside.js
zdlm: This is the same api in tradeshift-ui. I will show a message if the developer give us a string. if it is empty string, it will only show the spinner. If it is undefined, there will be nothing

packages/components/aside/src/aside.js
zdlm: Yes, it does stop. That is why I need to update it. I don't understand how does it been committed and merged.

zdlm:
> Looks ok for the aside but I checked https://tradeshift.github.io/elements/?path=/story/ts-spinner--default and I noticed few other things we need to address:
> 
> * The back color (the color of the complete circle) has to be @ts-color-gray-light with 50% opacity. The opacity ensures it looks ok against both white backgrounds and an overlay.
> * The small version needs to be 3px in thickness and the medium needs to be 4px
> * I think we need a knob to showcase it with a dark overlay (that'll default the color knob to white)

Cool. I will change it in anther PR

packages/components/aside/src/aside.js
esmailpour-hosein: Then maybe it would be better to just use boolean property for it, what do you think?

packages/components/aside/src/aside.js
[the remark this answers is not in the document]
esmailpour-hosein: using boolean property help with this as well

packages/components/aside/src/aside.js
zdlm: How could the developer pass a message if the property is boolean?

zdlm: For example: `data-busy="data loading..."`

esmailpour-hosein: Shouldn't it be a separate property? It makes it a bit confusing in my opinion but if you think this is a better, I am okay with it, we can refactor later if we decided other way

zdlm: It is not good idea to have more properties. And we have the same property in Tradeshift-ui. All developers use it like this way. They will get confused when we separate it to two.

esmailpour-hosein: I have a different opinion, and I think it make the api more clear what the property does. There are lots of new developers came and coming to Tradeshift and also 3rd party developers who are not used to the old api. If there's a room for improvement then the previous library should not stop us from doing it 👍

esmailpour-hosein: But as I mentioned above I'm okay with it if you think it's a better way to do it

packages/components/aside/stories/aside.stories.js
```diff
@@ -69,4 +69,23 @@ storiesOf('ts-aside', module)
 			`;
 		},
 		{ notes: readme }
+	)
+	.add(
```
esmailpour-hosein: Is it possible to just add the busy property to the above story?

PR #174, opened 2019-10-11
Add search icon and add circular background

esmailpour-hosein:
Screenshot 2019-10-11 at 14 03 17


packages/components/icon/src/icon.css
```diff
@@ -98,3 +98,12 @@ svg,
 :host([type='suggested']) svg {
 	color: var(--ts-color-purple);
 }
+
+:host([circular]) {
+	border-radius: 40em;
```
zdlm: What is the number ? 40em

esmailpour-hosein: Just a large number, Habit for making rounded corners for non square shapes, not necessary here since there are square already

PR #175, opened 2019-10-11
Update the style of spinner

zdlm:
The back color (the color of the complete circle) has to be @ts-color-gray-light with 50% opacity. The opacity ensures it looks ok against both white backgrounds and an overlay.
The small version needs to be 3px in thickness and the medium needs to be 4px


packages/core/src/vars.css
```diff
@@ -1,6 +1,5 @@
 :root {
 	/* Units........................................................ */
-
```
esmailpour-hosein: Was this reformat necessary?

packages/core/src/vars.css
```diff
@@ -150,32 +129,27 @@
...
 	--ts-progress-bar-height: 8px;
-
 	--ts-box-width-small: 280px;
 	--ts-box-width-medium: 360px;
-
 	--ts-shadow-color: var(--ts-color-slate-dark);
-
 	/* Spinner........................................................ */
+	--spinner-border-color: rgba(203, 215, 220, 0.5);
```
esmailpour-hosein: Can you add a comment to say what's the color name? /* gray-lightest*/ or something like that!

packages/core/src/vars.css
zdlm: I just run format in my IDE.

packages/core/src/vars.css
zdlm: 👍

packages/core/src/vars.css
esmailpour-hosein: Maybe we need to ad more stuff to the .editorconfig file for the future

PR #190, opened 2019-11-18
Lza/release

zdlm:

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,4 +1,4 @@
 {
 	"packages": ["packages/components/*", "packages/core"],
-	"version": "0.5.2"
+	"version": "0.5.3"
 }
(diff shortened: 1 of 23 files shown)


esmailpour-hosein:
Why you have 3 patch? 🤔

zdlm:
Because I made something wrong. I didn't create a branch at the first time. The opt was expired while publishing.

esmailpour-hosein:
For the next time that it happened you can just delete the tags and release on github by this commands:

```
# delete local tag '12345'
git tag -d 12345
# delete remote tag '12345' (eg, GitHub version too)
git push origin :refs/tags/12345
# alternative approach
git push --delete origin tagName
git tag -d tagName
```

and then deleting the branch and create another one 👍

zdlm:
Good job. I will put it into readme for publishing.

esmailpour-hosein:
Greate, We also need to fix the process, so it should do couple of other checks before just creating the tags so we won't end up in that situation

PR #191, opened 2019-11-18
Change readme of publishing

zdlm:

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 <h1 align="center">Elements</h1>
 
 <p align="center">
-  Reusable Tradeshift UI Components as Web Components 
+  Reusable Tradeshift UI Components as Web Components
     <a href="https://tradeshift.github.io/elements">
       https://tradeshift.github.io/elements
     </a>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -227,10 +227,21 @@ $ npm run component-gen
 
 ## ➤ How to release
 
-As mentioned before, we are using [lerna](https://github.com/lerna/lerna), to be able to publish you need enable your [two-factor authentication in npm](https://docs.npmjs.com/configuring-two-factor-authentication)
```
esmailpour-hosein: Why you removed the NPM OTP?

README.md
```diff
@@ -227,10 +227,21 @@ $ npm run component-gen
...
+
+- Create a branch for releasing。 `git branch release`
+- `git push release` and set the upstream.
+- `lerna publish`
+- If you have problems while releasing. The following may help
 
 ```shell
-$ NPM_CONFIG_OTP=123456 lerna publish`   # Push the packages to registry.npmjs.org
+# delete local tag '12345'
+git tag -d 12345
```
esmailpour-hosein: Can you change 12345 to the `v0.5.0` or something that shows the format of our release tag?

README.md
```diff
@@ -227,10 +227,21 @@ $ npm run component-gen
...
+# delete remote tag '12345' (eg, GitHub version too)
+git push origin :refs/tags/12345
+# alternative approach
+git push --delete origin tagName
+git tag -d tagName
```
esmailpour-hosein: One of the way to deleting the tag would be enough I think! Or maybe even moving this information to wiki page?
I think keeping the readme file as clean and lean as possible give it more chance to actually be read by the users not just scrolling through it

README.md
zdlm: They fixed right now. We don't need it anymore

README.md
zdlm: 👍

PR #197, opened 2019-11-27
Add initial implementation of a modal component

aleks-elkin:
Added a modal component according to the NAV-60 Jira ticket.
- two sizes: large (take the whole page with 20px padding) and medium (padding 25% from left and right, 12,5% from top and bottom)
- slots for note, main and footer.
- can be opened/closed via the "data-visible" property or the call functions open()/close()
- Emit events for start and finish of the animation of open/close.
 
Modal

P.S.: also in this PR changed one js function in the index.html to run without errors in IE11.

PR #198, opened 2019-11-27
vertically center the icon; border at the bottom; right-to-left knob

zdlm:

I looked over the Header element. Here are a few minor visual bugs:

vertically center the icon

border at the bottom

right-to-left knob

https://tradeshift.atlassian.net/browse/NAV-62

### DIFF
--- a/packages/components/header/src/header.css
+++ b/packages/components/header/src/header.css
@@ -7,6 +7,7 @@
 		width: 100vw;
 		height: var(--ts-unit-triple);
 		background-color: var(--ts-color-white);
+		border-bottom: var(--ts-border);
 		display: flex;
 		align-items: center;
 		flex-direction: row;
@@ -14,6 +15,7 @@
 
 	& .icon {
 		margin-left: var(--ts-unit-half);
+		block-size: var(--ts-unit-double);
 	}
 
 	& .title {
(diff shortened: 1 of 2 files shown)


packages/components/header/src/header.css
```diff
@@ -7,13 +7,15 @@
 		width: 100vw;
```
esmailpour-hosein: Using `vw` make it not useable in other components, it's the reason for horizontal scrollbar in storybook

packages/components/header/src/header.css
```diff
@@ -7,13 +7,15 @@
...
 		background-color: var(--ts-color-white);
+		border-bottom: var(--ts-border);
 		display: flex;
 		align-items: center;
 		flex-direction: row;
 	}
 
 	& .icon {
 		margin-left: var(--ts-unit-half);
+		block-size: var(--ts-unit-double);
```
esmailpour-hosein: What was the reason to choose `block-size`? IE and Edge not supported and I can't see any difference with or without them. 
https://developer.mozilla.org/en-US/docs/Web/CSS/block-size

packages/components/header/src/header.css
aleks-elkin: I address this in my pull request with modal component:
https://github.com/Tradeshift/elements/pull/197

packages/components/header/src/header.css
zdlm: The browser add some block-size make the icon not vertically center.

PR #215, opened 2019-12-19
Updated the storybook for table component.

aleks-elkin:
Refactored the existed `table` story to two separate stories:
- story with the custom render functions for some cells, but with fixed column and data values.
- story with the support of custom columns and data values, but, because of tool limitations (functions will disappear after JSON serialization to string) - without custom render functions.

### DIFF
--- a/packages/components/table/stories/table.stories.js
+++ b/packages/components/table/stories/table.stories.js
@@ -1,12 +1,110 @@
-import { storiesOf, html, select } from '@open-wc/demoing-storybook';
-import { withKnobs, array } from '@storybook/addon-knobs';
+import { storiesOf, html, select, text, array } from '@open-wc/demoing-storybook';
+import { withKnobs } from '@storybook/addon-knobs';
 
 import '@tradeshift/elements.table';
+import readme from '../README.md';
+
+/**
+ * Decodes the string with encoded html characters.
+ * @param {string} input
+ */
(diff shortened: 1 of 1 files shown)


packages/components/table/stories/table.stories.js
```diff
@@ -1,12 +1,110 @@
...
+		'default',
+		() => {
+			const columnsDefaultConfig = JSON.stringify([
+				{
+					property: 'name',
+					size: 'large',
+					value: 'Context'
+				},
+				{
+					visibility: 'desktop-only',
```
tradeshiftci: MINOR Define a constant instead of duplicating this literal 8 times. [rule](https://sonar.ts.sv/coding_rules#rule_key=javascript%3AS1192)

tradeshiftci:
SonarQube analysis reported 1 issue
* ![MINOR][MINOR] 1 minor

Watch the comments in this conversation to review them.

[MINOR]: https://sonarsource.github.io/sonar-github/severity-minor.png 'Severity: MINOR'

esmailpour-hosein:
The code is good but I don't think we should add this story to our storybook so people would think it's something that suggested to be used. I tried to find an ignore or disable kind of thing for a story but I couldn't find any. Maybe you can just rename the file so it doesn't match the regex for our stories.

PR #222, opened 2020-01-07
NAV-61 action button style for ts-button component

kirill-kolombet:
Made simple action button implementation **inside** _ts-button_.
Currently, we need only default _gray_ state. It's implemented as `type="text"` button with the icon specified. 

Normal:
Screenshot 2020-01-07 at 16 59 03

Hover:
Screenshot 2020-01-07 at 16 58 55

Pressed:
Screenshot 2020-01-07 at 16 58 59

### DIFF
--- a/index.html
+++ b/index.html
@@ -33,14 +33,14 @@
 				document.body.appendChild(coreEl);
 				coreEl.onload = function() {
 					var components = [
+						'button',
+						'button-group',
 						'root',
 						'list-item',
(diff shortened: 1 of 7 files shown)


kirill-kolombet:
Other states of the action button could be added with the new button design, which will require a serious revamping of the button code structure. I started to do it in the _kolombet/action-button_ branch, but it currently looks like overcomplicated.

PR #231, opened 2020-01-23
Lza/board

zdlm:
There are no buttons and tabs in this version of boar. 
I will give collaboration team to use this one first. 
Make buttons and tabs later

### DIFF
--- a/packages/components/aside/src/aside.css
+++ b/packages/components/aside/src/aside.css
@@ -11,7 +11,7 @@
 	min-height: 100vh;
 	width: var(--ts-aside);
 	background-color: var(--ts-color-white);
-	box-shadow: 0 0 3px var(--ts-shadow-color);
+	box-shadow: 0 0 3px var(--ts-color-slate-dark);
 	z-index: var(--ts-zindex-aside);
 }
 
--- a/packages/components/board/src/board.js
+++ b/packages/components/board/src/board.js
@@ -1,0 +1,44 @@
+import { TSElement, unsafeCSS, html, customElementDefineHelper } from '@tradeshift/elements';
+import css from './board.css';
+
+customElementDefineHelper(
+	'ts-board',
+	class extends TSElement {
+		static get styles() {
+			return [TSElement.styles, unsafeCSS(css)];
+		}
+
+		static get properties() {
+			return {
+				dir: { type: String, reflect: true },
+				title: { type: String, attribute: 'data-title' }
+			};
+		}
+
+		get direction() {
+			return this.dir ? this.dir : this.bodyDir;
+		}
+
+		get header() {
+			if (!this.title) {
+				return '';
+			}
+			return html`
+				<header class="board-header">
+					<div class="board-title">${this.title}</div>
+				</header>
+			`;
+		}
+
+		render() {
+			return html`
+				<div dir="${this.direction}" class="board-container">
+					${this.header}
+					<main class="board-main">
+						<slot name="main"></slot>
+					</main>
+				</div>
+			`;
+		}
+	}
+);
--- a/packages/components/board/stories/board.happo.js
+++ b/packages/components/board/stories/board.happo.js
@@ -1,0 +1,18 @@
+import { storiesOf } from '@open-wc/demoing-storybook';
+import '@tradeshift/elements';
+import '@tradeshift/elements.board';
+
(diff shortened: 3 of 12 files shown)


packages/components/aside/src/aside.css
```diff
@@ -11,7 +11,7 @@
 	min-height: 100vh;
 	width: var(--ts-aside);
 	background-color: var(--ts-color-white);
-	box-shadow: 0 0 3px var(--ts-shadow-color);
+	box-shadow: 0 0 3px var(--ts-color-slate-dark);
```
esmailpour-hosein: Shouldn't we change the `ts-shadow-color` to slate-dark?

packages/components/board/src/board.js
```diff
@@ -0,0 +1,49 @@
...
+		static get properties() {
+			return {
+				dir: { type: String, reflect: true },
+				title: { type: String, attribute: 'data-title' }
+			};
+		}
+
+		constructor() {
+			super();
+			this.title = '';
```
esmailpour-hosein: Why do we need this?

packages/components/board/src/board.js
```diff
@@ -0,0 +1,49 @@
...
+				</header>
+			`;
+		}
+
+		render() {
+			return html`
+				<div dir="${this.direction}" class="board-container">
+					${this.header}
+					<main class="board-main">
+						<slot name="main"></slot>
```
esmailpour-hosein: What if we don't put a name for it?

packages/components/board/stories/board.happo.js
```diff
@@ -0,0 +1,21 @@
...
+	const properties = {
+		dir: { rtl: 'rtl' },
+		title: { Title: 'Title' }
+	};
+
+	const options = {
+		columns: 5,
+		persistent_props: {
+			errorMessage: 'Failed message'
+		}
```
esmailpour-hosein: This need to be removed

packages/components/board/stories/board.stories.js
```diff
@@ -0,0 +1,16 @@
...
+
+import '@tradeshift/elements.board';
+
+storiesOf('ts-board', module)
+	.addDecorator(withKnobs)
+	.add('default', () => {
+		const title = text('My Board', 'Title');
+
+		return html`
+			<ts-board data-title="${title}">
```
esmailpour-hosein: Please also add `dir` prop 👍

packages/components/board/stories/board.stories.js
```diff
@@ -0,0 +1,16 @@
...
+				This my first board.
```
esmailpour-hosein: Can you please make this part also dynamic so people can provide `html` in the slot?

packages/core/src/vars.css
```diff
@@ -155,7 +155,7 @@
 	--ts-progress-bar-height: 8px;
 	--ts-box-width-small: 280px;
 	--ts-box-width-medium: 360px;
-	--ts-shadow-color: var(--ts-color-slate-dark);
+	--ts-shadow-color: var(--ts-color-gray-lightest);
```
esmailpour-hosein: Now I saw this one 🤔 Was the `aside` the only place we used this shadow-color?

packages/components/aside/src/aside.css
zdlm: The shadow in the aside is different with the current shadow in the board.
The slate-dark is only for aside, And the light-gray is for the common shadow.

packages/components/board/src/board.js
zdlm: 👍

packages/components/board/src/board.js
zdlm: I will add more slots. such as `tab`, `Button`.
Let's keep it right now.

packages/components/board/stories/board.happo.js
zdlm: 👍

packages/components/board/stories/board.stories.js
zdlm: 👍

packages/core/src/vars.css
zdlm: Yes. that is true

esmailpour-hosein:
Please check happo for IE11, It doesn't seems right.

PR #232, opened 2020-01-24
Added checks for initial update and re-render of search component

aleks-elkin:
Fixed in a `ts-search` component the emitting of the "idle" event on initial render and re-render with the same value.

### DIFF
--- a/packages/components/search/src/search.js
+++ b/packages/components/search/src/search.js
@@ -52,6 +52,14 @@ customElementDefineHelper(
 					this.dispatchIdleEvent = helpers.debounceEvent(() => this.dispatchCustomEvent('idle', this.value), newVal);
 					break;
 				case 'value':
+					if (!this.hasUpdated) {
+						// do not trigger events on initialization.
+						return;
+					}
+					if (newVal === oldVal) {
+						// do not trigger events on re-render with the same value (it happens in React).
+						return;
+					}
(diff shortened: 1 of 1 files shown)


packages/components/search/src/search.js
```diff
@@ -52,6 +52,14 @@ customElementDefineHelper(
 					this.dispatchIdleEvent = helpers.debounceEvent(() => this.dispatchCustomEvent('idle', this.value), newVal);
 					break;
 				case 'value':
+					if (!this.hasUpdated) {
```
esmailpour-hosein: I thought the problem was the debounce function, wasn't it?

aleks-elkin: No, after debug I found that the problem happens because the debounce function is triggered on component initialisation (when initial value changed from `null` to `''`). This should be fixed by first check.
Second check tries to solve another problem, which is related to React re-render on a state update. I created the small example: https://codesandbox.io/s/react-wc-search-component-91hfs

esmailpour-hosein: For the initial value check is it happening if developer set something or it's happening after we set it in the constructor?

aleks-elkin: It happens in both cases, the difference only in the initial value in the `newVal` - empty string(which is came from the constructor) or the value from developer.

PR #235, opened 2020-01-28
add sizes to ts-app-icon

teonaandreea:
add large size (60x60px) to ts-app-icon

### DIFF
--- a/packages/components/app-icon/src/app-icon.js
+++ b/packages/components/app-icon/src/app-icon.js
@@ -10,8 +10,7 @@ customElementDefineHelper(
 
 		static get properties() {
 			return {
-				// Use the type when have more sizes of app icon
-				type: { type: String, reflect: true },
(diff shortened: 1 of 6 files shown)


packages/components/app-icon/src/app-icon.js
```diff
@@ -10,21 +10,20 @@ customElementDefineHelper(
...
 		constructor() {
 			super();
-			this.type = '';
+			this.size = '';
 		}
 
 		render() {
 			return html`
-				<img .src="${this.src}" .alt="${this.alt}"></img>
+				<img .src="${this.src}" .alt="${this.alt}" size="${this.size}"></img>
```
esmailpour-hosein: Since you used the `:host([size='large'])` this size attribute and constructor are redundant. You should either remove them or change the css selector 👍

PR #241, opened 2020-03-10
Added a noPadding property to a modal

aleks-elkin:
Added a property to remove the default paddings of main content in a `ts-modal` component.

Modal no paddings

### DIFF
--- a/packages/components/modal/README.md
+++ b/packages/components/modal/README.md
@@ -33,12 +33,14 @@ table {
 
 ## ➤ Properties
 
-| Property | Attribute    | Type    | Default | Description                                                    |
-| -------- | ------------ | ------- | ------- | -------------------------------------------------------------- |
-| dir      | data-dir     | string  | ''      | Direction 'rtl' or 'ltr'                                       |
-| size     | data-size    | string  | 'large' | Size of the modal. Available two variants: 'large' or 'medium' |
-| title    | data-title   | string  | ''      | Modal header text                                              |
-| visible  | data-visible | boolean | false   | Show/hide the modal                                            |
+| Property   | Attribute    | Type    | Default | Description                                                    |
+| ---------- | ------------ | ------- | ------- | -------------------------------------------------------------- |
+| dir        | data-dir     | string  | ''      | Direction 'rtl' or 'ltr'                                       |
+| size       | data-size    | string  | 'large' | Size of the modal. Available two variants: 'large' or 'medium' |
+| title      | data-title   | string  | ''      | Modal header text                                              |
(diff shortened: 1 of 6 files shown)


packages/components/modal/README.md
```diff
@@ -33,12 +33,14 @@ table {
...
-| size     | data-size    | string  | 'large' | Size of the modal. Available two variants: 'large' or 'medium' |
-| title    | data-title   | string  | ''      | Modal header text                                              |
-| visible  | data-visible | boolean | false   | Show/hide the modal                                            |
+| Property   | Attribute    | Type    | Default | Description                                                    |
+| ---------- | ------------ | ------- | ------- | -------------------------------------------------------------- |
+| dir        | data-dir     | string  | ''      | Direction 'rtl' or 'ltr'                                       |
+| size       | data-size    | string  | 'large' | Size of the modal. Available two variants: 'large' or 'medium' |
+| title      | data-title   | string  | ''      | Modal header text                                              |
+| visible    | data-visible | boolean | false   | Show/hide the modal                                            |
+| hideHeader | hide-header  | boolean | false   | Show/hide thetitle of the modal                                |
```
esmailpour-hosein: Deliberately didn't add this to readme, to not encourage people for using it. Not sure if it's actually a good idea or not! You think we should add it?

aleks-elkin: I'm not sure that we should hide any public API that components have, because:
- I doubt that developers will use it without a real need (e.g. they usually have a design from UI/UX team),
- Most curious developers will find it anyway, because it is already in use in some apps.
I think we can create a warning mark for properties that we don't encourage to use.

packages/components/modal/src/modal.css
```diff
@@ -40,6 +40,10 @@ main {
 	padding: var(--ts-unit);
 }
 
+main.no-paddings {
```
esmailpour-hosein: We didn't decided before, how we want to write selectors for cases like this! Are we going to use the `host` like the following :
```css
:host([no-padding]) {
```
or do the way that you did. 

IMO using the host selector is a bit better and we can keep the property for documentation reason.

packages/components/modal/README.md
esmailpour-hosein: You are right but not all designs follow the principle to consider consistency, that much, and in this way we have the chance to at least stop some misuses.
Let's keep it then 👍

packages/components/modal/src/modal.css
aleks-elkin: Good point!
I changed to the `host` attribute selector, I also think this is a better approach for such cases.

PR #242, opened 2020-03-11
Tidy up in the package.json files

aleks-elkin:
- Removed the `gitHead` property from package.json files. I believe it was added because of the lerna issue https://github.com/lerna/lerna/issues/1880 , which is already fixed in our repo.
- Removed the `package-lock.json` files from leaf components. As I understand from lerna docs, we need only the root `package-lock.json` file.
Updated the deps of `ts-radio` and `ts-radio-group` components (I don't now why they not updated by lerna, I hope it is a one time issue).


esmailpour-hosein:
Should we add the components `package lock`s to . gitignore?

aleks-elkin:
> Should we add the components `package lock`s to . gitignore?

I'm not sure about it. As I can see in the lerna repo, they not have it in their .gitignore file.
Actually, i'm not sure how these package-lock.json files was created, may be it was an old issue, which already fixed.
Let's not add any changes to .gitignore now and see how it will go.

PR #257, opened 2020-04-07
Dispatch button-click custom event

### LINKED ISSUE #250: Add a click event for Button
Currently there is no any event for the button, so if button is disabled or busy - I have to check a button state manually every time.

esmailpour-hosein:
Before it relied on click event on the component which didn't respect
the disabled and busy state of the component.
Used button-click instead of click to not cause any breaking change

fix #250

### DIFF
--- a/packages/components/button/src/button.js
+++ b/packages/components/button/src/button.js
@@ -17,7 +17,7 @@ export class TSButton extends TSElement {
 		return {
 			type: { type: String, reflect: true },
 			size: { type: String, reflect: true },
-			busy: { type: String, reflect: true },
+			busy: { type: Boolean, reflect: true },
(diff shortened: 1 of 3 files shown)


packages/components/button/src/button.js
```diff
@@ -38,9 +38,15 @@ export class TSButton extends TSElement {
 		return colorBackgroundTypes.indexOf(this.type) > -1 ? 'inverted' : 'default';
 	}
 
+	clickHandler(event) {
+		if (!this.disabled) {
```
aleks-elkin, line 42: Just for clarity, please add an explicit check for the busy state.

esmailpour-hosein: done! 👍

PR #261, opened 2020-04-17
NAV-161  ts-tabs: Move content padding to ts-tab host

esmailpour-hosein:
To make it possible for developers to change the default padding of content in 
edge cases.

### DIFF
--- a/packages/components/tab/src/tab.css
+++ b/packages/components/tab/src/tab.css
@@ -3,3 +3,9 @@
 .tab-content:not([selected]) {
 	display: none;
 }
+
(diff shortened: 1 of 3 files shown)


packages/components/tab/src/tab.css
```diff
@@ -2,4 +2,11 @@
 
 .tab-content:not([selected]) {
 	display: none;
+	padding: var(--ts-unit);
```
aleks-elkin, line 5: I'm not sure, why you need this padding if the item is not displayed?

PR #263, opened 2020-04-17
Added release to github and credentials check

aleks-elkin:
- Added check of credentials in a release process. It checks that you logged to default registry, to Npm registry, and to Github registry.
- Splitted release to two phases:
  - Publish to Npm registry
  - publish to Github registry

Caveats: you need to delete the npm config setting `@tradeshift:registry https://npm.pkg.github.com/` because with this config setting it will always publish to Github registry. Added info about it to the Readme Release section.


README.md
```diff
@@ -237,7 +237,9 @@ We are using [lerna to publish](https://github.com/lerna/lerna/tree/master/comma
 
 - Create a branch for releasing。 `git branch release`
 - `git push release` and set the upstream.
+- `npm config delete @tradeshift:registry` delete the registry config so release could publish to different registries
 - `npm run release`
+- `npm config set @tradeshift:registry https://npm.pkg.github.com/` Set back npm to use github for @tradeshift packages
```
esmailpour-hosein, line 242: I was wondering is it possible to merge these step to release script?

esmailpour-hosein:
@aleks-elkin please put your information about why you decided to do it in this way and not using gitub action, if there's any specific reason 🙏

README.md
aleks-elkin: I think it would be better to do in a separate shell script, because it would be nice to check this value first, change it only if it exists, and return to the initial state after release.

aleks-elkin:
> @aleks-elkin please put your information about why you decided to do it in this way and not using gitub action, if there's any specific reason 🙏

I decided to implement a manual release first, because I have some additional information to discuss, and I want to do this discussion before implement an automation of releases.

PR #268, opened 2020-04-27
Added a checkbox-changed event to checkbox

aleks-elkin:
Added a handler for click event to update the `checked` property status of checkbox.

### DIFF
--- a/packages/components/checkbox/src/checkbox.js
+++ b/packages/components/checkbox/src/checkbox.js
@@ -26,12 +26,21 @@ export class TSCheckbox extends TSElement {
 		return this.dir || this.bodyDir;
 	}
 
+	onClick(e) {
+		e.stopPropagation();
+		if (this.disabled) {
+			return;
+		}
+		this.checked = !this.checked;
+	}
+
 	render() {
 		return html`
 			<label class="checkbox-container">
 				${this.label}
 				<input
 					type="checkbox"
+					@click="${this.onClick}"
 					.name="${this.name}"
 					.value="${this.value}"
 					?checked="${this.checked}"


packages/components/checkbox/src/checkbox.js
```diff
@@ -26,12 +29,22 @@ export class TSCheckbox extends TSElement {
 		return this.dir || this.bodyDir;
 	}
 
+	onClick(e) {
+		e.stopPropagation();
+		if (this.disabled) {
+			return;
+		}
+		this.checked = !this.checked;
+		this.dispatchCustomEvent(customEventNames.CHECKBOX_CHANGED, this.checked);
```
esmailpour-hosein, line 38: I think it's a good idea to send the original event object beside the new checked value.

esmailpour-hosein: Like this one : https://github.com/Tradeshift/elements/blob/1a3c45c188d00116446f0ce7bb2d14c8983d78a2/packages/components/file-uploader-input/src/file-uploader-input.js#L147

aleks-elkin:
> What do you think if we just let them know they can use MutationObserver? 🤔
> https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver
@esmailpour-hosein 

Well, the general idea looks good, but we still need to fix the component: currently the `checked` attribute is not reflected properly because of behavior of the `checked` attribute of the HTML input element:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#checked

> A Boolean attribute indicating whether or not this checkbox is checked by default (when the page loads). It does not indicate whether this checkbox is currently checked: if the checkbox’s state is changed, this content attribute does not reflect the change.

esmailpour-hosein:
Yes, you are right, I test it out and works fine, but don't know what would be missed in this way.
Other than that this component, missing the light DOM input, which make `name` and `value` attributes redundant. We should take care of that as well in future.

aleks-elkin:
> Yes, you are right, I test it out and works fine, but don't know what would be missed in this way.
> Other than that this component, missing the light DOM input, which make `name` and `value` attributes redundant. We should take care of that as well in future.

Yes, it is one of the tasks for future improvement: add support for forms. But currently developers cannot use it effectively because `checked` status not updated after user interactions, so I wanted to solve this one.

PR #269, opened 2020-04-28
Changed the z-index of modal to be below aside.

aleks-elkin:
Changed z-index of the modal to be below the aside.
The same behaviour we have in TS-UI.
https://github.com/Tradeshift/tradeshift-ui/blob/485f88919cec5ee6dcdcc47f9bb0cd78a1bc224d/src/runtime/less/ts-modals.less#L3

https://github.com/Tradeshift/tradeshift-ui/blob/626bbd3357c057608ec40cc8779ce0be76ab04dd/src/runtime/less/ts-variables.less#L41

Screenshot 2020-04-28 at 16 12 07

### DIFF
--- a/packages/components/cover/src/cover.css
+++ b/packages/components/cover/src/cover.css
@@ -45,3 +45,10 @@
 		z-index: var(--ts-zindex-spin-cover);
 	}
 }
+
+:host(.ts-modal-cover) {
+	& > div {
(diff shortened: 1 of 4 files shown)


esmailpour-hosein:
Can you please share a screenshot of the result 🙏

PR #273, opened 2020-05-01
Aditional status classes

teonaandreea:
The status label is missing 2 states - primary(blue) and the error state used to be orange (corresponding to warning) instead of red.
The naming of the status was not conforming to the naming conventions we use in other components.

This PR introduces the following naming + colors:

- ok -> changed to "success"
- note -> changed to "neutral"
- error -> changed color to red instead of orange
- warning (new) -> orange (the colors that previously belonged to error)
- primary -> blue

### DIFF
--- a/packages/components/basic-table/stories/basic-table.happo.js
+++ b/packages/components/basic-table/stories/basic-table.happo.js
@@ -101,7 +101,7 @@ storiesOf('ts-basic-table', module).add('test', () => {
 						mobileDescription: 'Tradeshift | 5 participants',
 						type: 'document'
 					},
-					type: { status: 'ok', text: 'External' },
+					type: { status: 'success', text: 'External' },
 					participants: 100,
 					newActivity:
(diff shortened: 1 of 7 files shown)


packages/components/basic-table/stories/basic-table.happo.js
```diff
@@ -101,7 +101,7 @@ storiesOf('ts-basic-table', module).add('test', () => {
 						mobileDescription: 'Tradeshift | 5 participants',
 						type: 'document'
 					},
-					type: { status: 'ok', text: 'External' },
+					type: { status: 'success', text: 'External' },
```
aleks-elkin, line 104: Could you please change the type status "error" to "warning" in the next row object? So it will look the same as it was and there will be no diffs in happo.

teonaandreea: sure 👍

PR #274, opened 2020-05-01
fix: board: added default bg and changed shadow

aleks-elkin:
- Added default white background for a board element. It is defined on `:host` level, so developers can change it easily.
- Background for header will be always white and text-color will be black so the host settings will not affect the header.
- Changed box-shadow to the same we have in ts-ui.
- Fixed the story to correctly render content knob.

Some notes: for the best rendering I had to put shadow and border-radius on `:host` level and add the `overflow:hidden` to clip the inside content by corner borders. Downside of this approach is that developers can mess up with that.

Default board:
Default board
Board and parent with custom background:
Board on custom bg

Board with custom background for content and parent with custom background:
board on custom bg with custom bg

### DIFF
--- a/packages/components/board/src/board.css
+++ b/packages/components/board/src/board.css
@@ -1,13 +1,18 @@
 /* General........................................................ */
 
-.board-container {
-	border: var(--ts-border-width) var(--ts-border-type) var(--ts-border-color-default);
-	box-shadow: 1px 2px var(--ts-shadow-color);
+:host {
+	background-color: var(--ts-color-white);
+	border: var(--ts-border);
 	border-radius: var(--ts-radius);
+	box-shadow: var(--ts-board-shadow);
+	display: block;
+	overflow: hidden;
 }
 
 .board-header {
-	border-bottom: var(--ts-border-width) var(--ts-border-type) var(--ts-border-color-default);
+	background-color: var(--ts-color-white);
+	border-bottom: var(--ts-border);
+	color: var(--ts-color-black);
 	padding: var(--ts-unit-half) var(--ts-unit);
 	font-size: var(--ts--fontsize);
 	font-weight: var(--ts-fontweight-semibold);
--- a/packages/components/board/stories/board.stories.js
+++ b/packages/components/board/stories/board.stories.js
@@ -1,18 +1,22 @@
 import { storiesOf, html } from '@open-wc/demoing-storybook';
-import { withKnobs, text } from '@storybook/addon-knobs';
+import { withKnobs, text, radios } from '@storybook/addon-knobs';
 
 import '@tradeshift/elements.board';
 
 storiesOf('ts-board', module)
 	.addDecorator(withKnobs)
-	.add('default', () => {
-		const dir = text('ltr', 'Dir');
-		const title = text('My Board', 'Title');
-		const content = text('Put any html inside the board', 'Content');
-
-		return html`
-			<ts-board dir="${dir}" data-title="${title}">
-				${content}
-			</ts-board>
-		`;
-	});
+	.add(
+		'default',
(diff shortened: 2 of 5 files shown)


packages/components/board/stories/board.stories.js
```diff
@@ -5,14 +5,19 @@ import '@tradeshift/elements.board';
...
-		return html`
-			<ts-board dir="${dir}" data-title="${title}">
-				${content}
-			</ts-board>
-		`;
-	});
+	.add(
+		'default',
+		() => {
+			const dir = text('ltr', 'Dir');
```
esmailpour-hosein, line 11: Can you please change this to a radio? We should refactor all elements and make them consistent.

packages/components/board/src/board.css
```diff
@@ -1,13 +1,18 @@
 /* General........................................................ */
 
-.board-container {
+:host {
+	background-color: var(--ts-color-white);
 	border: var(--ts-border-width) var(--ts-border-type) var(--ts-border-color-default);
-	box-shadow: 1px 2px var(--ts-shadow-color);
 	border-radius: var(--ts-radius);
+	box-shadow: 0 2px 2px rgba(0, 0, 0, 0.1), 0 2px 8px rgba(0, 0, 0, 0.05);
```
esmailpour-hosein, line 7: I don't think the border, border-radius and box-shadow should be override-able. Should we bring back the board-container class name and have them there?

packages/components/board/stories/board.stories.js
```diff
@@ -5,14 +5,19 @@ import '@tradeshift/elements.board';
...
+			const title = text('My Board', 'Title');
```
esmailpour-hosein, line 12: I think the label and value are in wrong order. Another think for sake of consistency and clarity, I think we should use the attribute for the label, so in here it should be `data-title` instead of 'Title'. What do you think?

packages/components/board/stories/board.stories.js
```diff
@@ -5,14 +5,19 @@ import '@tradeshift/elements.board';
...
+			const mainDiv = document.createElement('div');
+			mainDiv.slot = 'main';
+			mainDiv.innerHTML = text('Put any html inside the board', '<h3>Content</h3>');
```
esmailpour-hosein, line 15: Same here we should have a consistent way to demonstrate slots in stories. I don't have any specific idea right now, just another thing to discuss in tech talk

packages/components/board/src/board.css
esmailpour-hosein: And maybe you can reuse the style in vars.css or if it's not available create a new one.

aleks-elkin: if we put it inside the host, it will be clipped by "overflow:hidden" css rule. Without the "overflow:hidden" the inside content will not have rounded borders.
Board without overflow hidden

packages/components/board/stories/board.stories.js
aleks-elkin: Yeah, we definitely should make the `slot` feature clear to developers.
Actually, in board we have only one slot, so it could be a default one, without names, so everything that placed inside the `ts-board` will be placed in this slot. But it would be breaking changes.

esmailpour-hosein: Yes, one concern that we had before was for developers who are not familiar with the concept of slots, if we make changes and add another slot for example, it would be confusing for them to have the default slot and next to that named slot, probably that's the reason Leo used named slot. But I think we should use default slot and have a clear documentation with examples in those cases.

esmailpour-hosein:
@aleks-elkin Check the board without title in IE11, It need to be fixed I think

PR #276, opened 2020-05-05
fix(checkbox): improved layout. Fixed a checked state management

### LINKED ISSUE #275: Checkbox without label renders incorrectly
With label:
checkbox with label

Without label:
checkbox without label

Checkbox should keep block size without text in label.

aleks-elkin:
- Fixed #275 

- Fixed a `checked` state management. Before there could be unexpected behaviour:
  - render checkbox in `checked` state
  - User unchecks it.
  - set `checked` attribute from code.
  - Now it will have a `checked` attribute but will render unchecked. If user will click on it, a `checked` attribute removed but checkbox rendered in checked state.

- Added a few knobs to a story.

### DIFF
--- a/packages/components/checkbox/src/checkbox.js
+++ b/packages/components/checkbox/src/checkbox.js
@@ -36,18 +36,17 @@ export class TSCheckbox extends TSElement {
 
 	render() {
 		return html`
-			<label class="checkbox-container">
-				${this.label}
+			<div dir="${this.dir}" class="checkbox-container" @click="${this.onClick}">
 				<input
 					type="checkbox"
-					@click="${this.onClick}"
 					.name="${this.name}"
 					.value="${this.value}"
 					?checked="${this.checked}"
 					?disabled="${this.disabled}"
 				/>
-				<span></span>
-			</label>
+				<div class="checkbox"></div>
+				${this.label}
+			</div>
 		`;
 	}
 }
--- a/packages/components/checkbox/src/checkbox.css
+++ b/packages/components/checkbox/src/checkbox.css
@@ -1,9 +1,7 @@
 /* General........................................................ */
 
 .checkbox-container {
-	display: block;
-	position: relative;
-	padding-left: var(--ts-unit-plus);
+	display: flex;
 	margin: var(--ts-unit-half);
 	cursor: pointer;
 	font-size: var(--ts-fontsize);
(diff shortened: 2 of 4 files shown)


packages/components/checkbox/src/checkbox.js
```diff
@@ -36,18 +36,17 @@ export class TSCheckbox extends TSElement {
...
-					@click="${this.onClick}"
 					.name="${this.name}"
 					.value="${this.value}"
 					?checked="${this.checked}"
 					?disabled="${this.disabled}"
 				/>
-				<span></span>
-			</label>
+				<div class="checkbox"></div>
+				${this.label}
```
esmailpour-hosein, line 48: Since we are not forcing to have text for the `label`, I think it does make sense to have slot for it to be customizable

aleks-elkin: I'm not sure what do you mean by this, can you explain a bit more? Currently we don't have any slots and do not support any customisation for a label.

esmailpour-hosein: Since developers can just don't provide text for the label and put it next some other stuff like an image or icon, then maybe we can replace attribute with slot in this case, to let them put icon, image, etc. in the slot, so they can have the benefit of triggering checkbox by click on the label as well.

esmailpour-hosein: Right now, they can just put an empty checkbox and next to it, use an image, and with some custom js get the similar functionality

aleks-elkin: Ok, I see now. But how checkbox should be vertically aligned with this content? Also, I think we need to discuss it with UI/UX team first, do we want to provide such functionality.
And I suggest to implement it in a separate PR, since it will be a feature, not a fix.

esmailpour-hosein: Sure we can do it in a separate PR 👍 
But it's not a feature to provide, it's already achievable, only in a bit hacky way.

esmailpour-hosein:
Please check the IE11 screenshots

packages/components/checkbox/src/checkbox.js
aleks-elkin: I mean, that In terms of standard `ts-checkbox` functionality it will be considered as feature, because right now you cannot achieve this behaviour declaratively, without additional code on application side.
Of course you can implement many additional features on application side, but it doesn't mean that it's `native` implementation should be considered as a fix :)

esmailpour-hosein:
There is an issue in IE11 for the Happo test which is not reproducible in browserstack. We'll take care of the issue later if anyone encountered this, to unblock this PR.

PR #282, opened 2020-05-06
fix(checkbox): returned a click event propagation

aleks-elkin:
In the previous commit, I accidentally added the `event.stopPropagation()` method call for the `click` event, which is why all listeners outside the element stopped receiving this event.
Now I fix this.

### DIFF
--- a/packages/components/checkbox/src/checkbox.js
+++ b/packages/components/checkbox/src/checkbox.js
@@ -26,8 +26,7 @@ export class TSCheckbox extends TSElement {
 		return this.dir || this.bodyDir;
 	}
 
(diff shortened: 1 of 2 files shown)


packages/components/checkbox/src/checkbox.js
```diff
@@ -27,7 +27,6 @@ export class TSCheckbox extends TSElement {
 	}
 
 	onClick(e) {
```
esmailpour-hosein, line 29: Can you please remove `e`?

aleks-elkin: Good catch!
Removed it.

PR #289, opened 2020-06-15
Automate the properties table generation in readme files

esmailpour-hosein:
Added scripts to automate the process of reading properties, events and slots data from src file and
generate readme file tables based on them.

There are couple of steps in the script:
1. It tries to read the current data from the readme file and generate a properties.json inside docs/readme directory of each component. (only for properties)
2. It reads the data from js file of the component and generate another properties/events/slots.json inside docs/src directory of each component. 
3. Uses Plop readme action to generate and update the current readme file of the component, from readme template under plop-templates/readme/README.md
4. Within the previous task, in data function, it merge the information from source file and readme to update the readme file

Also added the prerelease script to package.json to run this script before each release. Still need some other script to commit and push README files.

Next step would be generating the types file as well.


packages/components/app-icon/src/app-icon.js
```diff
@@ -8,8 +8,11 @@ export class TSIcon extends TSElement {
 
 	static get properties() {
 		return {
+			/** Size of the app icon, available variants: ''(default), 'large' */
```
aleks-elkin, line 11: Do you think we can use [this](https://github.com/jsdoc/jsdoc/issues/629#issue-31314808) approach to document limited options for params? E.g.:
```js
/**
* Size of the app icon.
* @type {('default'|'large')}
*/
```
My intention with this proposal is to have a standard way to document limited options, because I think this would help with generating `d.ts` files later.

packages/components/aside/src/aside.js
```diff
@@ -3,49 +3,37 @@ import css from './aside.css';
 import '@tradeshift/elements.button';
 import '@tradeshift/elements.cover';
 import '@tradeshift/elements.spinner';
-import { customEventNames } from './utils';
```
aleks-elkin, line 6: I personally would like to have these event names as an exportable enum, because IMHO it improves the dev experience and makes a code less error-prone, when you use enum values instead of plain string values. And it brings additional value for typescript projects - with enum they will have a compile-time check that the event name actually exists.
Example of how I think it can be used in a separate project:
https://github.com/aleks-elkin/react-typescript-web-components/blob/master/src/components/ElementsModal.tsx#L24

Also, with separate enum we could add a JSDoc comment on top of each item, which could be useful for documentation as well.

aleks-elkin (commented):
Overall looks awesome!
I have just a few comments that I would like to discuss first before approve it.

packages/components/app-icon/src/app-icon.js
esmailpour-hosein: I agree, it should be like that, I had something similar in mind, but deliberately didn't included it in this PR. I think we should create the guideline for consistency and include when and how these enums should be used. If you think it's better to do in this PR I will do it.

packages/components/aside/src/aside.js
esmailpour-hosein: I also like them and think this pattern can be useful but after introducing auto-generated prop from resource, I believe they can be actually generated to help other developers for using the event name, slot name and etc. I think removing them is helping with keeping everything in one place and make it simpler for people to contribute and maintain. These were (and some still are) inconsistent part of the library that can be fixed by removing unnecessary convention, so we don't miss it in review process.

I think we should keep all the related information about a slot or event in one place and document it in the same place, but generate those files for developers using the library, to don't loose the benefits of them. This should be handled in the same PR that we'll generate the types files. What do you think?

packages/components/app-icon/src/app-icon.js
aleks-elkin: Ok, sounds good to do it in a separate PR. I just want to be sure that we are on the same page :)

packages/components/aside/src/aside.js
aleks-elkin: I'm a bit afraid that source code could became a mess when we have documentation comments in the middle of the implementation, but it is just a guess. Let's try to do it in your way and see how it will look.

Another corner case to think about: what to do when we dispatch the same event in several places? Should we document all of them or just the first one?
- If only one would be enough, but then a developer have to keep a track when each event is introduced for the first time.
- if all should be documented, we potentially could have an inconsistency because people make mistakes and could forget to update all docs.

esmailpour-hosein: I agree that it can get really messy and personally don't like code comments in general, no matter what they've being used for, but there's a trade-off that I think at least for now it worth the try. Let's get back to this after updating all descriptions and other info in those comments.
Great point! 🤔 
I also believe it would rare cases that happen in the library and especially in smaller/simpler components that won't happen. My suggestion for those cases is a disable next line comment (eslint, ... ). What do you say? I can add it to this PR.

aleks-elkin: Ok, let's try your approach to put comments in implementation and see how it will work out :)

I don't understand your suggestion about eslint, can you elaborate a bit on it?
To clarify, my concern is what to do when we have such situation:

```js
class TSComponent extends TSElement {
  ...
   someFunc () {
    this.dispatchCustomEvent('some-event-name', {data: "someData"});
  }
  ...
   anotherFunc () {
    this.dispatchCustomEvent('some-event-name', {data: "someData"});
  }
}
```
We have two functions that dispatch the same event.
Should the developer put JSDoc comment about event only in first case or put it on both events?
And what to do if after refactoring we swap these functions and `anotherFunc` became higher than `someFunc`?

It is not a blocker for this PR, just a corner case that I could imagine.

esmailpour-hosein: It should be only one of them I would say, but not important which one.
I think the script should get improved and cover these cases and if there are two instances with jsdoc, throw an error to let developers know that there's another place that this event being documented.

aleks-elkin (approved):

scripts/from-src/get-properties.js
```diff
@@ -0,0 +1,74 @@
...
+	let defaults = defaultsContent.split('constructor(){super()')[1];
+	if (defaults === undefined) {
+		return;
+	}
+	defaults = defaults.split('}')[0];
+	defaults = defaults.split('this.').map(data => data.split('='));
+	defaults.shift();
+
+	defaults = defaults.filter(item => item.length === 2);
+	return defaults.reduce((o, k) => ((o[k[0]] = k[1]), o), {});
```
aleks-elkin, line 44: I do not completely understand, what is happening in this line of code. Could you add some comments, or use more descriptive names and probably split it to separate functions?

scripts/from-src/index.js
```diff
@@ -0,0 +1,12 @@
+const getSrcData = require('./get-data-from-src');
+const saveDataFromSrc = require('./save-data-from-src');
+const { compStateLogger } = require('../helpers');
+const xa = require('xa');
```
aleks-elkin, line 4: I don't see how this is used here. Consider to remove it or add a usage.

scripts/from-src/get-slots-data.js
```diff
@@ -0,0 +1,26 @@
+module.exports = function getSlotData(componentName, fileContent) {
+	const slots = [];
+	const contentSplitBySlot = fileContent.split('<slot');
+
+	contentSplitBySlot.forEach((content, index) => {
+		let name,
+			description = '';
```
aleks-elkin, line 7: Something wrong with formatting :)

aleks-elkin (approved):
Overall seems nice, just a few comments.
Also, check the warnings from SonarCube and fix the ones that seems reasonable to you :)

scripts/from-src/get-properties.js
esmailpour-hosein: I refactored it to be clear.

packages/components/search/README.md
```diff
@@ -26,29 +26,30 @@
...
 
-| Event Name | Payload | Description                                                         |
-| ---------- | ------- | ------------------------------------------------------------------- |
-| change     | string  | Emitted on every user's change in a search input                    |
-| idle       | string  | Emitted when the user not change input value for a provided timeout |
-| search     | string  | Emitted when the user press the 'Enter' key                         |
+| Name   | Description                                                         | Payload            |
+| ------ | ------------------------------------------------------------------- | ------------------ |
+| idle   | Emitted when the user not change input value for a provided timeout | search input value |
+| idle   |                                                                     |                    |
```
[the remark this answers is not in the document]
esmailpour-hosein: This was something that could be fixed by the code refactor, I hope that we can dodge that for now.

PR #304, opened 2020-11-25
fix: Do not preselect the first radio without checked attr

isprogfun:
According to MDN having none radios checked in the beginning is a valid situation which is not the case for our elements https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#Value. I've added negative index to handle this. It'll work with keyboard navigation as well. First `toggle` won't do anything without having a radioElement and it's the case with the negative index.

### DIFF
--- a/packages/components/radio-group/src/radio-group.js
+++ b/packages/components/radio-group/src/radio-group.js
@@ -75,7 +75,7 @@ export class TSRadioGroup extends TSElement {
 	handleSlotChange() {
 		this.radios = this.radioElements;
 
-		this.index = 0;
+		this.index = -1;
 		for (let i = 0; i < this.radios.length; i++) {
 			if (this.radios[i].checked) {
 				this.index = i;
--- a/packages/components/radio-group/src/utils/helper-functions.js
+++ b/packages/components/radio-group/src/utils/helper-functions.js
@@ -25,7 +25,7 @@ export function getNextElementIndex(direction, index, collectionLength) {
 			index++;
 			return index;
 		case directionTypes.BACKWARD:
-			if (index === 0) {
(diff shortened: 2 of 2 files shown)


esmailpour-hosein:
Thanks for contribution @isprogfun 
I think we should not make it the default behaviour but an option. Both because it's not the most used case and also I don't want to cause breaking change just yet. So please add another property for it to disable this default behaviour.
We also need @mihneaz approval before be able to merge this.

mihneaz:
I hope I understood this right: we are automatically checking the first radio input in any group and we'd like to disable that functionality when needed. 
If that's the case, yes, we'd like to have that functionality but not as the default behaviour.

isprogfun:
> So please add another property for it to disable this default behaviour.

Ok. What name would you suggest for the property then? How should it behave when you have this property and also have a radio with the `checked` parameter? What would be the priority then?

esmailpour-hosein:
My suggestion is `disable-default-initial-select`, know it's ugly but can't think of any more descriptive name, if you or @aleks-elkin have a better suggestion, go for it 👍 
It should respect the `checked` property of the radio item, otherwise developers would have to remove this if they show the form in edit mode.

esmailpour-hosein (changes requested):
My suggestion is `disable-default-initial-select`, know it's ugly but can't think of any more descriptive name, if you or @aleks-elkin have a better suggestion, go for it 👍
It should respect the `checked` property of the radio item, otherwise developers would have to remove this if they show the form in edit mode.

isprogfun:
Hi, in the new design there is no need to have this component so let's not introduce new options and keep the existing behaviour

PR #317, opened 2021-02-24
[PEAA-57] fix(pager): fixed a css media query for mobile devices

aleks-elkin:
Fixed a mistake in css rule that blocks a transition of ts-pager to mobile state. Now it should work fine.
Pager mobile

### DIFF
--- a/packages/components/pager/src/pager.css
+++ b/packages/components/pager/src/pager.css
@@ -123,7 +123,7 @@ li {
 	}
 }
 
-@media screen and (env(--mobile-screen-breakpoint)) {
+@media screen and (max-width: env(--mobile-screen-breakpoint)) {
 	.container {
 		margin: var(--ts-unit-half);
 	}


packages/components/pager/src/pager.css
```diff
@@ -123,7 +123,7 @@ li {
 	}
 }
 
-@media screen and (env(--mobile-screen-breakpoint)) {
+@media screen and (max-width: env(--mobile-screen-breakpoint)) {
```
esmailpour-hosein, line 126: 🙈

esmailpour-hosein (approved):

packages/components/pager/src/pager.css
aleks-elkin: The issue was introduced in April 2020, but reported just today 😄

esmailpour-hosein: It was actually one of the issues that Mihnea mentioned about the components that we have :D 
I think we have a ticket for it that can get closed now 👍  
Another sign why we should prioritize improving our testing. 
I created some stuff in hackathon around using the codeceptjs for visual regression testing which can help us to get rid of the cost limitation of happo and do more tests in different screen size and seperate scenarios, including hover state. I'll bring it up during the planning.

PR #320, opened 2021-03-25
feat(popover): PEAA-316. Added an initial implementation for popover.

aleks-elkin:
Added an initial implementation of the popover element.

There are two ways to close the popover: by pressing the `X` button on top right of the popover or via API by setting `opened` property to `false`.

If the anchor element is defined, it will try to position the popover relative to this anchor. Developer should specify `placement` to define where the popover should be rendered.

If the anchor is not specified, popover will be placed absolutely, based on the values in `positionLeft` and `positionTop`, which accepts any valid css syntax, e.g. `50px`, `50%`, `calc(100%-50px)`.

Screenshot:
Popover

### DIFF
--- a/packages/components/popover/src/popover.css
+++ b/packages/components/popover/src/popover.css
@@ -1,0 +1,51 @@
+/* General........................................................ */
+:host {
+	display: none;
+	outline: none;
+	pointer-events: none;
+}
+
+:host([opened]) {
+	display: block;
+	z-index: var(--ts-zindex-max);
+	position: fixed;
+	top: 0;
+	left: 0;
+	width: 100%;
+	height: 100%;
+}
+
+.popover-container {
+	background: var(--ts-color-black);
+	color: var(--ts-color-white);
+	font-size: var(--ts-fontsize-small);
+	padding: var(--ts-unit-half);
+	border-radius: var(--ts-radius);
+	position: absolute;
+	pointer-events: all;
+	max-width: 375px;
+}
+
+.popover-header {
+	height: var(--ts-unit);
+	display: flex;
+	justify-content: space-between;
+	font-size: var(--ts-fontsize-small);
+
+	& .popover-title {
+		flex: 1 1 var(--ts-unit-half);
+		font-weight: var(--ts-fontweight-semibold);
+	}
+
+	& .popover-close {
+		flex: 0 0 var(--ts-unit-half);
+		margin-left: var(--ts-unit-half);
+		cursor: pointer;
+	}
+}
+
+.popover-footer {
+	display: flex;
+	justify-content: flex-end;
+	font-size: var(--ts-fontsize-small);
+}
--- a/packages/components/popover/src/popover.js
+++ b/packages/components/popover/src/popover.js
@@ -1,0 +1,190 @@
+import { TSElement, unsafeCSS, html, customElementDefineHelper } from '@tradeshift/elements';
+import '@tradeshift/elements.icon';
+import css from './popover.css';
+
+/**
+ * Placement relative to the anchor element.
+ */
+export const Placement = {
+	TOP_LEFT: 'topLeft',
+	TOP_RIGHT: 'topRight',
+	BOTTOM_LEFT: 'bottomLeft',
+	BOTTOM_RIGHT: 'bottomRight'
+};
+
+export class TSPopover extends TSElement {
+	static get styles() {
+		return [TSElement.styles, unsafeCSS(css)];
+	}
+
+	static get properties() {
+		return {
+			/** Is the popover visible or not */
+			opened: { type: Boolean, reflect: true },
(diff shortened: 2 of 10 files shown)


packages/components/popover/src/popover.css
```diff
@@ -0,0 +1,51 @@
...
+
+.popover-header {
+	height: var(--ts-unit);
+	display: flex;
+	justify-content: space-between;
+	font-size: var(--ts-fontsize-small);
+
+	& .popover-title {
+		flex: 1 1 var(--ts-unit-half);
+		font-weight: 600;
```
esmailpour-hosein, line 37: Can we use a css variable here?

packages/components/popover/src/popover.js
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	static get properties() {
+		return {
+			/** Is the popover visible or not */
+			opened: { type: Boolean, reflect: true },
+			/** Position on X axis relative to the anchor. Could be 'left', 'right' */
+			positionX: { type: String, attribute: 'position-x', reflect: true },
+			/** Position on Y axis relative to the anchor. Could be 'top', 'bottom' */
+			positionY: { type: String, attribute: 'position-y', reflect: true },
```
esmailpour-hosein, line 33: It's my personal preference but I think having a single attr for positioning/placement is cleaner and easier to handle both in our code and usage of the component.  `topRight`, `topLeft`, `bottomRight`, `bottomLeft` and we can add more options to it later.
I suggest change the name with `placement` as well to not cause confusion with position-left/top attrs.

packages/components/popover/src/popover.js
```diff
@@ -0,0 +1,185 @@
...
+			/** Text in the title */
+			header: { type: String, reflect: true },
+			/** Anchor element for relative positioning. If you need to position absolutely, leave this empty */
+			anchor: { type: String, reflect: true },
+			/** Left offset when popover is positioned absolutely. Use any valid css syntax for the 'left' property */
+			positionLeft: { type: String, attribute: 'position-left', reflect: true },
+			/** Top offset when popover is positioned absolutely. Use any valid css syntax for the 'top' property */
+			positionTop: { type: String, attribute: 'position-top', reflect: true }
```
esmailpour-hosein, line 41: Not quite sure if it's better to keep these separated or as one attr.

packages/components/popover/src/popover.js
```diff
@@ -0,0 +1,185 @@
...
+		};
+	}
+
+	constructor() {
+		super();
+		this.opened = false;
+	}
+
+	onClose() {
+		this.opened = false;
```
esmailpour-hosein, line 51: If we are going with [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) it's okay not emit an event for this. But we should either create a helper for it or provide an example with explanation in our documentations, I think it's out of scope of  this PR.

packages/components/popover/src/popover.js
```diff
@@ -0,0 +1,185 @@
...
+	stopEventBubbling(e) {
+		e.stopPropagation();
+	}
+
+	attributeChangedCallback(name, oldVal, newVal) {
+		super.attributeChangedCallback(name, oldVal, newVal);
+		switch (name) {
+			case 'opened':
+				if (newVal !== null) {
+					window.customElements.whenDefined('ts-popover').then(() => {
```
esmailpour-hosein, line 63: What's the reason for it? Can this be triggered before the component being defined? 🤔

packages/components/popover/src/popover.js
```diff
@@ -0,0 +1,185 @@
...
+		Object.assign(container.style, position);
+	}
+
+	getByIdFromParents($elem, query) {
+		// If a shadow root doesn't exist we don't continue the traversal
+		if ($elem.shadowRoot == null) {
+			return null;
+		}
+
+		// Grab the rood node and query it
```
esmailpour-hosein, line 158: typo :)

esmailpour-hosein (approved):
Great job 👍 
Just some comments. Please add the happo and storybook stuff 🙏

packages/components/popover/src/popover.js
aleks-elkin: Great suggestion! I also did not like the possible confusion between `positionX` and `positionLeft`. I think it looks better now.

packages/components/popover/src/popover.js
aleks-elkin: Well, have them separate is easier for implementation 😄

packages/components/popover/src/popover.js
aleks-elkin: Good point, I added a custom event for it.

packages/components/popover/src/popover.js
aleks-elkin: Because we are trying to get the internal DOM node `container`, but it would be unavailable until the element is defined and template is processed. This callback helps with this problem.

PR #327, opened 2021-03-30
[PEA-150] Fix missing styling for checkbox component

esmailpour-hosein:
- Added missing styling for focus state and fix the hover state styling
- Make the hover and focus state of components more consistent
- Added slot for checkbox component to make the label customizable


packages/components/checkbox/src/checkbox.js
```diff
@@ -50,7 +50,8 @@ export class TSCheckbox extends TSElement {
 					?disabled="${this.disabled}"
 				/>
 				<div class="checkbox"></div>
-				${this.label}
+				<!-- To customized checkbox label (links, ...) -->
+				<slot> ${this.label} </slot>
```
aleks-elkin, line 54: We need to fix the story, because there is a whitespace and label do not render.
also, maybe we should somehow document, that data-label will not work if you have any whitespaces and special characters in your template, e.g. this examples will not render the label:
```html
<ts-checkbox data-label="Label"> </ts-checkbox>

<ts-checkbox data-label="Label">
</ts-checkbox>
```

aleks-elkin (approved):
Looks good!
I have just a few minor comments.

aleks-elkin (approved):

PR #329, opened 2021-04-21
feat(select): PEAA-168: Implement a single level select component.

aleks-elkin:
Initial implementation of the single level select component with multi-selection support.

Screenshots (the strange white/grey background is not because of component):
Closed dropdown

## Single select
Empty:
Single select

With value:
Single with value

## Multiselect
One selected item:
Single item

Several selected items:
Several items

Opened with several items:
Multi opened

Opened with uncommitted changes:
Multi opened with changes


packages/components/select/src/select.js
```diff
@@ -0,0 +1,328 @@
...
+		if (!this.opened) {
+			return;
+		}
+		requestAnimationFrame(() => {
+			const selectInput = this.shadowRoot.getElementById('selectInput');
+			const dropdown = this.shadowRoot.getElementById('dropdown');
+			const listContainer = this.shadowRoot.getElementById('listContainer');
+			if (!dropdown) {
+				// it could happen when element is just initialised but not rendered yet.
+				return;
```
esmailpour-hosein, line 147: nit, I think it would be better to move the selectInput and listContainer after the `if`.

packages/components/select/src/select.js
```diff
@@ -0,0 +1,328 @@
...
+			} else {
+				position.top = height + 10 + 'px';
+				maxHeight = paddingBottom > MAX_HEIGHT ? MAX_HEIGHT : paddingBottom;
+			}
+
+			maxHeight -= APPLY_CONTAINER_HEIGHT;
+			maxHeight += 'px';
+
+			Object.assign(dropdown.style, position);
+			Object.assign(listContainer.style, { maxHeight });
```
esmailpour-hosein, line 168: I couldn't figure it out why but the max-height doesn't match the design as I can see. The design itself has inconsistency, the top part says it has max-height of 600px but in the examples when there's a scrollbar, you can see its height is less than 400px. We should check this with @mihneaz

index.html
```diff
@@ -45,8 +45,48 @@
 	</head>
 
 	<body>
-		<article>
-			<ts-button>Button</ts-button>
+		<article style="height: 300px">
+			<ts-button id="button">Button</ts-button>
+			<br />
```
esmailpour-hosein, line 50: Please remove these

index.html
```diff
@@ -45,8 +45,48 @@
...
+			<br />
+			<br />
+			<br />
+			<ts-select id="select" multiselect placeholder="Choose your destiny"></ts-select>
 		</article>
+		<script>
+			(function () {
+				const select = document.getElementById('select');
+				select.items = [
```
esmailpour-hosein, line 70: I'm not quite sure if we should prefer properties over slots for cases like this or not. This is still something that we have to discuss and find a clear answer for. If we ever find time for that :D 
Maybe in future we can improve this component to let them add options with slots as well.

esmailpour-hosein (commented):
Great job! 👍  Just couple of comments, I don't think we would need a big change. 

Another small thing is the padding of the dropdown, which doesn't match with the design.

Just another thing that is not directly related to this component, but the list-item/menu component doesn't match with design library. In the designs it has rounded borders which is not the same in ts-ui and Elements. We should fix one of them. Also the selected state of the menu, is not the same as selected state of the select component.

esmailpour-hosein:
The focus state of the input doesn't match with the design.

esmailpour-hosein:
Disable and readonly states are also missing from the design

packages/components/select/src/select.js
```diff
@@ -0,0 +1,328 @@
...
+			</ts-button-group>
+		`;
+	}
+
+	/** @private */
+	getPlaceholder() {
+		if (this.selected.length === 0) {
+			return this.placeholder || '';
+		}
+		if (this.selected.length === 1) {
```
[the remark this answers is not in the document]
aleks-elkin: Good catch! I made changes so it should be showed as a normal text. Tried to make it inside the `update` callback, but the element start behaving strange (e.g. when you open it does not clear the input but do not use this text for filtering).
So I decided to remove the `update` callback and introduce the function `updateInputValue` and call it explicitly on every close of the dropdown.
I tested it locally in chrome, it works fine for me.

packages/components/select/src/select.js
```diff
@@ -0,0 +1,328 @@
...
+			const selectedItem = this.items.filter(item => item.id === this.selected[0])[0];
+			if (selectedItem) {
+				return selectedItem.title;
+			} else {
+				return this.placeholder || '';
+			}
+		}
+
+		return 'Selected ' + this.selected.length;
```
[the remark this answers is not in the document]
aleks-elkin: I wanted to do it, but completely forgot about it. Thanks for the catching it!

packages/components/select/README.md
```diff
@@ -35,17 +35,18 @@
 
 | Property | Attribute | Type | Default | Description |
 | --- | --- | --- | --- | --- |
-| dir | dir | String | ltr | Direction of the component 'rtl' or 'ltr'. |
+| dir | dir | String | ltr | Does component has uncommited changes or not. |
```
esmailpour-hosein, line 38: What? 👀  Is it the readme script doing this? :D

packages/components/select/src/select.js
```diff
@@ -29,6 +48,8 @@ export class TSSelect extends TSElement {
 			selected: { type: Array, reflect: true },
 			/** Default placeholder when there is no selection. */
 			placeholder: { type: String, reflect: true },
+			/** Translated messages for the user locale */
```
esmailpour-hosein, line 51: nit, just ranting, we should think of a way to include the translation default object in the readme, by modifying the update-readme script 🤔 
Although at first we should come to an agreement in the guideline about translations.

packages/components/select/src/select.js
```diff
@@ -0,0 +1,328 @@
...
+		this.opened = !this.opened;
+	}
+
+	open() {
+		this.opened = true;
+	}
+
+	firstUpdated() {
+		window.addEventListener('resize', this.resizeListener);
+		window.addEventListener('mousedown', this.clickOutsideListener);
```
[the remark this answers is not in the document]
aleks-elkin: You are right, previous implementation was more like proof of concept, just to make sure that it will work.
Moved it to the `attributeChangedCallback` and subscribe to these events only when it is open.
Also I want to keep the `disconnectedCallback` because I could imagine the case where the dropdown is open (and we have subscribed listeners) but app decide to remove it from the DOM (e.g. after some slow AJAX response).

Regarding make one toggle function with optional `open` parameter - I prefer to have several functions with different names, rather than one function with different behavior depending on the parameter.

packages/components/select/src/select.js
```diff
@@ -0,0 +1,328 @@
...
+	 * @param {Event} event
+	 */
+	handleOutsideClick(event) {
+		if (!this.opened) {
+			return;
+		}
+		if (!('composedPath' in event)) {
+			return;
+		}
+		const paths = event.composedPath();
```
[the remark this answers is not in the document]
aleks-elkin: I tested in IE11 and it is working, so apparently w e have a polyfill for it. But I did not investigate where it came from. Main suspects: babel in rollup config (it has `target: 'ie11'`) and polyfills' bundle of WebComponents.

packages/components/select/src/select.js
aleks-elkin: I did not notice the section above the `Select component` in Figma, so I choose this `400px` just as a reasonable value for me. Changed to `600px`.

index.html
aleks-elkin: Yeah, I also think, that we could provide just a slot and allow developers put their select items. But I think it would be tricky with multi-level component, plus I'm not sure what benefits it will give developers if we will provide slots instead of API, because at the end they will have the same data model (array of objects with names and ids).
Anyway, we should discuss it first :)

packages/components/select/src/select.js
```diff
@@ -67,23 +81,18 @@ export class TSSelect extends TSElement {
...
 
-	firstUpdated() {
-		window.addEventListener('resize', this.resizeListener);
-		window.addEventListener('mousedown', this.clickOutsideListener);
-	}
-
 	disconnectedCallback() {
 		window.removeEventListener('resize', this.resizeListener);
-		window.removeEventListener('mousedown', this.clickOutsideListener);
+		window.removeEventListener('click', this.clickOutsideListener);
```
esmailpour-hosein, line 86: I can't find where you add this listener.

esmailpour-hosein (commented):
Just some comments, I'm checking out locally to test it out.

packages/components/select/src/select.js
aleks-elkin: I agree, we should provide the default object and available strings (with description where they are used) in the README, but we should have a guideline first to have it in consistent way (plus be able to automate it with scripts).

packages/components/select/src/select.js
aleks-elkin: oops, this is actually a way to fix this issue:
when user clicks on the button outside the dropdown, this listener get triggered first and close the dropdown. Because of that the toggle function will always open the dropdown, because it will be closed at the time of onClick callback gets called.
To fix this I changed `mousedown` to `click` plus I also had to call a `stopPropagation` of the `click` event in button. I update the PR with this fix.

packages/components/select/README.md
aleks-elkin: Yep, completely forgot to tell you this. For some reason it takes the description from the `dirty` property instead of `dir`. I did not investigate it, but this issue is reproducible if you will call `npm run update-doc-src-from-readme` and then `npm run update-readme`. You can investigate it in this branch if you have time and desire to do it :)

esmailpour-hosein (approved):

PR #333, opened 2021-05-05
Add text field component

esmailpour-hosein:


packages/components/input/src/input.css
```diff
@@ -2,36 +2,115 @@
...
 /* Disable........................................................ */
 
 ::slotted([disabled]) {
 	color: var(--ts-disabled-color);
 }
+
+/* With icons ........................................................ */
+
+.icon-right,
+.icon-left {
```
aleks-elkin, line 79: I suggest to change naming to "icon-start" and "icon-end" so it would be more clear in case of RTL.

packages/components/input/src/input.js
```diff
@@ -8,16 +9,38 @@ export class TSInput extends TSElement {
...
+			/** Disable state of the input */
+			disabled: { type: Boolean, reflect: true },
+			/** Readonly state of the input */
+			readonly: { type: Boolean, reflect: true },
+			/** Direction 'rtl' or 'ltr' */
+			dir: { type: String, reflect: true },
+			/** Icon that appears at the end part of the input. Readonly and disabled state will show a lock icon instead. */
+			iconRight: { type: String, reflect: true, attribute: 'icon-right' },
+			/** Icon that appears at the beginning of the input */
+			iconLeft: { type: String, reflect: true, attribute: 'icon-left' }
```
aleks-elkin, line 23: I suggest to change naming to "iconStart" and "iconEnd" so it would be more clear in case of RTL.

packages/components/input/src/input.js
```diff
@@ -8,16 +9,38 @@ export class TSInput extends TSElement {
...
 		};
 	}
 
+	get direction() {
+		return this.dir ? this.dir : this.bodyDir;
+	}
+
+	get iconRightHtml() {
+		const iconRight = this.readonly || this.disabled ? 'lock' : this.iconRight;
+		return iconRight ? html` <ts-icon class="icon-right" icon=${iconRight} size="large" type="action"></ts-icon> ` : '';
```
aleks-elkin, line 33: nit: why there are spaces at start and end? Consider to remove it.

packages/components/input/src/input.js
```diff
@@ -26,28 +49,52 @@ export class TSInput extends TSElement {
...
 		}
 		return elements[0];
 	}
 
 	render() {
 		return html`
-			<div>
+			<div dir=${this.direction}>
+				${this.iconLeft &&
+				html` <ts-icon class="icon-left" icon=${this.iconLeft} size="large" type="action"></ts-icon> `}
```
aleks-elkin, line 94: nit: why there are spaces at start and end in string literal? Consider to remove it.

packages/components/text-field/src/text-field.css
```diff
@@ -0,0 +1,34 @@
...
+	text-transform: uppercase;
+}
+
+:host([disabled]) {
+	& label {
+		color: var(--ts-disabled-color);
+	}
+}
+
+.required-astrix {
```
aleks-elkin, line 32: typo: should be 'asterisk' instead of 'astrix' :)
Please, change it here and in .js file as well.

packages/components/text-field/src/text-field.js
```diff
@@ -0,0 +1,163 @@
...
+
+		const messages = this.hasError ? this.errorMessages : this.helpTextMessages;
+		const type = this.hasError ? 'error' : undefined;
+		if (messages.length > 1) {
+			const title = this.hasError ? this.errorTitle : this.helpTextTitle;
+			return html`
+				<ts-help-text
+					type=${type}
+					title=${title}
+					messages=${JSON.stringify(messages)}
```
aleks-elkin, line 106: You can use a property instead of attribute and do not stringify value. e.g.:
```
.messages=${messages}
```

packages/components/text-field/src/text-field.js
```diff
@@ -0,0 +1,163 @@
...
+					?rtl=${this.direction === 'rtl'}
+				></ts-help-text>
+			`;
+		}
+
+		return html` <ts-help-text
+			type=${type}
+			messages=${JSON.stringify(messages)}
```
aleks-elkin, line 114: You can use a property instead of attribute and do not stringify value.

packages/components/text-field/src/text-field.js
```diff
@@ -0,0 +1,163 @@
...
+			/** Placeholder of the text field */
+			placeholder: { type: String, reflect: true },
+			/** Array of messages to pass to helpt-text component. See help-text component for more info  */
+			helpTextMessages: { type: Array, reflect: true, attribute: 'help-text-messages' },
+			/** If you have more than one help text message , you should pass a title to it. See help-text component for more info  */
+			helpTextTitle: { type: String, reflect: true, attribute: 'help-text-title' },
+			/** Error messages to show underneath of the input when it has error */
+			errorMessages: { type: Array, reflect: true, attribute: 'error-messages' },
+			/** Error title, if there are more than one error message */
+			errorTitle: { type: Array, reflect: true, attribute: 'error-title' },
```
aleks-elkin, line 39: I'm pretty sure it's type should be a String instead of Array :)

aleks-elkin (commented):
Overall looks great!
I have just a few minor comments, hopefully it will not be a big change.

packages/components/input/src/input.css
esmailpour-hosein: I agree, I was between right/left and pre/post just went for it for sake of simpler to understand. I'll change it to the start/end 👍

packages/components/input/src/input.js
esmailpour-hosein: our format command is stupid  with lit-html stuff 😕

packages/components/text-field/src/text-field.css
esmailpour-hosein: If you say so ;) 
image

packages/components/text-field/src/text-field.js
esmailpour-hosein: Ahh, you are right! finally a usecase for this `.` properties 😅

packages/components/text-field/src/text-field.js
esmailpour-hosein: You are absolutely right :) Wondering how it did work in the storybook screenshot 🤔

aleks-elkin: In Happo stories there is only one error message, so `error-title` is not used there.

esmailpour-hosein: I meant the PR description screenshot: 
https://github.com/Tradeshift/elements/pull/333#issue-630368033

packages/components/text-field/src/text-field.js
```diff
@@ -0,0 +1,177 @@
...
+			originalEvent: event
+		});
+	}
+
+	get labelHtml() {
+		return html`
+			<label for=${this.id}>
+				<!-- If you want to have custom html in label, you can use this slot  -->
+				<slot name="label"><span>${this.label}</span></slot>
+				${this.required ? html` <span title="Required" class="required-asterisk">*</span> ` : ''}
```
aleks-elkin, line 94: Please, remove leftover spaces in string literal (we definitely have to tweak our format tool to not leave it behind).

aleks-elkin (approved):
Looks awesome! :shipit:

aleks-elkin (approved):

esmailpour-hosein:
Happo screenshot for text-field & help-text in ie11 is not similar to what I can see in browserstack and based on the fact that we are close to deprecate ie11, I'm not going to spend more time on this for now. We'll get back to it and fix it if needed:
Screenshot 2021-05-10 at 15 27 30

PR #353, opened 2021-07-07
Migrated to husky v7 and added a commitlint with GHA check for each PR

aleks-elkin:
- Migrated husky from v4 to v7 (there are some breaking changes).
- Added a commitlint package to validate commit messages with conventional commits standard. It will be used later to make automatic semantic releases based on conventional changelog with lerna. Documentation about the format: https://www.conventionalcommits.org/en/v1.0.0/
- Added a GHA to automatically check commit messages in each PR.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run update-readme && npm run format && npx lint-staged && npm run check-deps
--- a/package.json
+++ b/package.json
@@ -133,6 +133,8 @@
   "devDependencies": {
     "@babel/core": "^7.12.10",
     "@babel/preset-env": "^7.12.11",
+    "@commitlint/cli": "^12.1.4",
+    "@commitlint/config-conventional": "^12.1.4",
     "@rollup/plugin-babel": "^5.2.2",
     "@rollup/plugin-commonjs": "^17.0.0",
     "@rollup/plugin-json": "^4.1.0",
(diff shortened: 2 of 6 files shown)


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run update-readme && npm run format && npx lint-staged && npm run check-deps
```
wejendorp, line 4: Wont this be a bit much? Or is it quite fast to run those "all files" hooks?

package.json
```diff
@@ -67,12 +67,8 @@
...
     "update-readme": "cross-env-shell node ./scripts/update-readme-files.js",
     "add-desc-to-src": "cross-env-shell node ./scripts/add-description-to-src.js",
-    "watch": "cross-env-shell lerna exec --parallel -- rollup -w -c=`pwd`/rollup.config.js"
-  },
-  "husky": {
-    "hooks": {
-      "pre-commit": "npm run update-readme && lint-staged && npm run check-deps"
-    }
+    "watch": "cross-env-shell lerna exec --parallel -- rollup -w -c=`pwd`/rollup.config.js",
+    "prepare": "husky install"
```
wejendorp, line 71: Why did we choose `prepare`? Seems it's only relevant on `postinstall`, not on publish etc

wejendorp (approved):
Consider the comments, but nice, go ahead :)

.husky/pre-commit
aleks-elkin: Well, all of them where here before, I just move them to new config file for husky.
On my machine it took several seconds to complete. It is noticeable, but not annoying, IMO.

package.json
aleks-elkin: It was done by husky migration tool. I think here you could find the explanation behind this decision:
https://github.com/typicode/husky/issues/884

wejendorp: Great, thanks :)

> Unlike postinstall, the prepare command does not run when a NPM package is installed as a dependency in another project.

That's reason enough :+1:

PR #359, opened 2021-07-26
fix(select): make search in a select case-insensitive

aleks-elkin:
Make a search in a select component not case-sensitive by calling `toLowerCase()` and comparing results.


esmailpour-hosein (approved):

esmailpour-hosein:
🚀

PR #360, opened 2021-07-26
chore(release): v0.27.3

aleks-elkin:


packages/components/app-icon/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Change Log
+
+All notable changes to this project will be documented in this file. See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+## [0.27.3](https://github.com/Tradeshift/elements/compare/v0.27.2...v0.27.3) (2021-07-26)
+
+**Note:** Version bump only for package @tradeshift/elements.app-icon
```
esmailpour-hosein, line 7: Can we customize this message?

esmailpour-hosein (approved):

packages/components/app-icon/CHANGELOG.md
aleks-elkin: Good question:
It is still an open issue on lerna's issue tracker: https://github.com/lerna/lerna/issues/2188
But there is a some third-party package to remove this: https://codsen.com/os/lerna-clean-changelogs-cli/
Do you think we should try this package?

esmailpour-hosein: Thanks for the information. 
Not that important for me, just wondering if we could point out what was changed and use better wording about this is only a version bump :)

aleks-elkin: @esmailpour-hosein I updated the PR with result of the package run that I mentioned in previous comment. Do you think it looks better?

esmailpour-hosein: I kinda prefer the previous one with the note, but I'm okay with moving on with this one.

esmailpour-hosein (approved):

packages/components/app-icon/CHANGELOG.md
aleks-elkin: Yes, I also like the previous one better because it is explicit why the version was bumped.
Regarding the customisation of the message itself it looks like the message is hardcoded inside the lerna: https://github.com/lerna/lerna/blob/main/core/conventional-commits/lib/make-bump-only-filter.js#L17

So the only option is to reimplement changelog generation on our side which is I don't want to do :)

I will revert changes to the previous one and merge it.

PR #362, opened 2021-07-28
ci: add a GHA to check npm credentials

aleks-elkin:
For some reason GHA for publish is not able to publish new packages to NPM.
We need to debug the access to npm in GHA, so I added a Github action to check npm credentials on PR.


callebjorkell (approved):
The commit message/PR description is missing a "why", but other than that it's fine I suppose :smile:

PR #363, opened 2021-07-28
chore(release): v0.27.4

aleks-elkin:
Created another release with no actual changes, just to test our publish GHA.


wejendorp (approved):
These don't look like no changes, if there are changelogs? :shrug:

aleks-elkin:
> These don't look like no changes, if there are changelogs? 🤷

Well, in normal situation it should be run only when there are any actual changes, so I decided to not introduce too much logic with checks for any changes that need to be published.

PR #369, opened 2021-08-31
Add an overlay component to render elements on top of everything

aleks-elkin:
Added a new component: `overlay`. It should be added to `document.body` and it will render it's content on top of everything and relative to the provided anchor element. It could be used in components with dropdown functionality like `select`, `action-select` and `date-picker` as well as `popover` (and maybe `tooltip`).

Also made a migration of `select` component to use this `overlay` component to render it's dropdown part. Also made a code refactoring of `select` by moving the dropdown part with it's selection logic to a separate component `select-menu`.


packages/components/overlay/src/overlay.js
```diff
@@ -0,0 +1,197 @@
...
+		const paddingTop = top - HALF_UNIT;
+		const paddingBottom = innerHeight - bottom - HALF_UNIT;
+		if (paddingBottom < contentHeight && paddingBottom < paddingTop) {
+			// When we don't have enough space below and have more space above, we put dropdown above.
+			maxHeight = top > MAX_HEIGHT ? MAX_HEIGHT : paddingTop;
+			if (maxHeight < contentHeight) {
+				// maxHeight will be set as a new height of a content container after position update
+				contentHeight = maxHeight;
+			}
+			verticalPosition.top = top - height - contentHeight + HALF_UNIT + 'px';
```
callebjorkell, line 161: We were talking about that this is weird that it works (or then line 163), it seems like this should not work with the height being included in both top and bottom calculations.

callebjorkell (approved):
Very nice walk through of the code, and I like it all! :smile:

packages/components/overlay/src/overlay.js
aleks-elkin: I found the problem:
the `top` value actually equals the length between top window border and top border of the anchor: https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect

So, the calculation for the bottom positioning (line 163) is correct. But the calculation for the top positioning was wrong and looks fine just because height was near 20px, and `-height + HALF_UNIT` was almost equal to `-HALF_UNIT`. So the correct one should be: (top) - (height of the content) - (gap between anchor and dropdown == HALF_UNIT).

PR #371, opened 2021-09-01
Update deps and bump core-js version in storybook babel config

aleks-elkin:
Bumped outdated dependencies (fixing the `tar` vulnerability among others from npm audit) and updated the `core-js` version in storybook configuration to the latest one.


callebjorkell (approved):
I see from the lines that this is fine.

PR #374, opened 2021-09-16
feat(select): (PDV-1405) select items without confirmation apply button

aleris:
adds option to select (and implicitly to select-menu) to allow direct
items selections without a separate apply button

https://user-images.githubusercontent.com/1011245/133579804-0960d730-5770-4b43-861f-40e9de4e87d1.mov

### DIFF
--- a/index.html
+++ b/index.html
@@ -33,7 +33,7 @@
 				coreEl.setAttribute('src', '/packages/core/lib/core.umd.js');
 				document.body.appendChild(coreEl);
 				coreEl.onload = function () {
-					var components = ['button'];
+					var components = ['icon', 'typography', 'button', 'overlay', 'select-menu', 'list-item', 'select'];
 					components.forEach(function (component) {
 						var el = document.createElement('script');
 						el.setAttribute('src', '/packages/components/' + component + '/lib/' + component + '.umd.js');
@@ -49,5 +49,20 @@
 		<article>
 			<ts-button>Button</ts-button>
 		</article>
+		<article>
+			<ts-select
+				id="s1"
+				multiselect
+				no-apply-button
+				items='[{"id":0,"title":"Item 1"},{"id":1,"title":"Item 2"},{"id":2,"title":"Item 3"}]'
+			/>
+		</article>
+		<script>
+			window.addEventListener('load', () => {
+				document.getElementById('s1').addEventListener('select-changed', e => {
+					console.log('select-changed', e);
+				});
+			});
+		</script>
 	</body>
 </html>
--- a/packages/components/select-menu/src/select-menu.js
+++ b/packages/components/select-menu/src/select-menu.js
@@ -21,6 +21,9 @@ export class TSSelectMenu extends TSElement {
 			items: { type: Array, reflect: true },
 			/** Allow users to select several options or not. */
 			multiselect: { type: Boolean, reflect: true },
+			/** Do not show the apply button and directly emit select-menu-changed when the selection changes.
+			 * Only affects the behaviour when multiselect is enabled, for single selection this is the default behavior. */
+			noApplyButton: { type: Boolean, reflect: true, attribute: 'no-apply-button' },
 			/** List of selected items' ids */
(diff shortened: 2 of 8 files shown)


index.html
```diff
@@ -33,7 +33,7 @@
 				coreEl.setAttribute('src', '/packages/core/lib/core.umd.js');
 				document.body.appendChild(coreEl);
 				coreEl.onload = function () {
-					var components = ['button'];
+					var components = ['icon', 'typography', 'button', 'overlay', 'select-menu', 'list-item', 'select'];
```
aleris, line 36: this changes are not needed, but maybe they can remain here as in another example of using another more complex component

esmailpour-hosein: I'm working on an improvement for this, hopefully each component would have their own development file with the proper setup to make the process easier.

esmailpour-hosein:
For future reference: https://tradeshift.slack.com/archives/C02TX8JJ7/p1631798756019800?thread_ts=1631777845.017200&cid=C02TX8JJ7

index.html
aleris: would be nice. let me know if you want to keep this changes or should I revert it

packages/components/select-menu/src/select-menu.js
```diff
@@ -21,6 +21,9 @@ export class TSSelectMenu extends TSElement {
 			items: { type: Array, reflect: true },
 			/** Allow users to select several options or not. */
 			multiselect: { type: Boolean, reflect: true },
+			/** Do not show the apply button and directly rise select-menu-changed when the selection changes.
+			 * Only affects the behaviour when multiselect is enabled, for single selection this is the default behavior. */
+			noApplyButton: { type: Boolean, reflect: true },
```
esmailpour-hosein, line 26: Please add the `attribute` property for it. 
attribute: 'no-apply-button'

I'm not a fan of the name but it's really hard to find any better name for property of this feature. `autoSelect` `disableApplySelect`, .... I don't think these are any better, but maybe you had some other options in mind to share?

aleris: not very happy with the name also, but do not have any other great ideas, 
hideApplyPanel
noApplyPanel
doNotShowApply
also no hard opinion on it

packages/components/select-menu/src/select-menu.js
```diff
@@ -21,6 +21,9 @@ export class TSSelectMenu extends TSElement {
 			items: { type: Array, reflect: true },
 			/** Allow users to select several options or not. */
 			multiselect: { type: Boolean, reflect: true },
+			/** Do not show the apply button and directly rise select-menu-changed when the selection changes.
```
esmailpour-hosein, line 24: nit: I think `emit` is a bit better word instead of `rise`

packages/components/select-menu/src/select-menu.js
esmailpour-hosein: I think the one that you chose is most clear option for now. Hopefully we can kill the apply button functionality all together in next iterations! 
But please do add the attribute for it to be kebab-case.

esmailpour-hosein (commented):
LGTM. @aleks-elkin can you please take a look as well?

packages/components/select/stories/select.stories.js
```diff
@@ -33,6 +33,7 @@ export const Default = () => {
 	const disabled = boolean('disabled', false);
 	const opened = boolean('opened', false);
 	const multiselect = boolean('multiselect', false);
+	const noApplyButton = boolean('multiselect', false);
```
aleks-elkin, line 36: nit: please, rename from `multiselect` to `no apply button` or similar.

aleks-elkin (approved):
Looks good, just one minor comment.

Thank you for contribution!

packages/components/select/stories/select.stories.js
aleris: good catch

aleris:
can you help with making Happo run?

PR #377, opened 2021-09-27
fix(select): no-apply-button attribute wrongly set on select-menu

aleris:
when changed the attribute to snake case no-apply-button, it was
not propagated correctly to the select-menu used in the select

see also: PDV-1407


esmailpour-hosein (commented):
I think you should also change the noApplyButton in the storybook of select. 
https://github.com/Tradeshift/elements/blob/c0596dd82eecc9e178b54c0d7ddb8c6501cc56ae/packages/components/select/stories/select.stories.js#L56

esmailpour-hosein (approved):

PR #380, opened 2021-09-30
feat(date-picker): add date-picker component

esmailpour-hosein:
Screenshot 2021-10-01 at 10 29 53
Screenshot 2021-10-01 at 10 29 43
Screenshot 2021-10-01 at 10 29 32


packages/components/date-picker/stories/date-picker.happo.js
```diff
@@ -0,0 +1,189 @@
...
+				<p>Date picker</p>
+				<div style="margin-bottom: 400px">
+					Closed:
+					<ts-date-picker></ts-date-picker>
+				</div>
+
+				<div>
+					Opened: (deliberately testing showing current month with today dot, it needs manual approval every time it
+					runs in a different day)
+					<ts-date-picker opened></ts-date-picker>
```
esmailpour-hosein, line 32: I'm not sure if it would be better to mock Date, to not have it fail? or is it okay to accept the change everytime?

packages/components/date-picker/src/date-picker.js
```diff
@@ -0,0 +1,281 @@
...
+			/** INTERNAL */
+			selectedDateObj: { type: String, reflect: false },
+			/** Label of the date picker. */
+			label: { type: String, reflect: true },
+			/** Is it possible to deselect the date? You will be able to change the date still. */
+			deselectable: { type: Boolean, reflect: true },
+			/** Is the date picker is disabled? */
+			disabled: { type: Boolean, reflect: false },
+			/** Is the date picker is readonly? */
+			readonly: { type: Boolean, reflect: false },
```
aleks-elkin, line 49: I not see any mockups for readonly state. As a matter of fact, in readonly state your input is not disabled and you still able to type in or open overlay and change the date, it just renders a lock icon on the right side of the input :)

packages/components/date-picker/stories/date-picker.happo.js
aleks-elkin: I personally prefer to mock dates, because tests that are always failing could distract when something else really fails. If you really want to test that it correctly renders on different dates, i think it is better to create several tests that have mocks with different dates :)
Because this date comes from outside of the component, I think it doesn't matter, was the date a true real one or we mocked it.

packages/components/date-picker-overlay/src/date-picker-overlay.js
```diff
@@ -0,0 +1,198 @@
...
+		const classNames = [
+			'day-cell',
+			inPageMonth ? 'day-current-month' : '',
+			day.isToday ? 'today' : '',
+			day.isSelected ? 'selected' : '',
+			day.isDisabled ? 'disabled' : ''
+		].join(' ');
+		return html`<span @click="${() => !day.isDisabled && this.selectDate(day.date)}" class="${classNames}"
+			>${day.date.getDate()}</span
+		>`;
```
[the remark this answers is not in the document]
esmailpour-hosein: Unfortunately this is the prettier doing this to make sure there's no unwanted spaces in the span tag.

packages/components/date-picker-overlay/src/date-picker-overlay.js
```diff
@@ -0,0 +1,198 @@
...
+				${this.deselectable &&
+				html`
+					<ts-button
+						class="deselect-button"
+						@button-click="${this.deselect}"
+						?disabled="${!this.selectedDate}"
+						type="secondary"
+						full-width
+						>Deselect</ts-button
+					>
```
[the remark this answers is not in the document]
esmailpour-hosein: Same as above, result of running `format` script.

packages/components/date-picker/src/date-picker.js
esmailpour-hosein: Besides that, what a description :)) `Is the date picker is readonly?` :))))
I'll fix it and add the test and knob to storybook. Nice catch.

packages/components/date-picker/src/date-picker.js
```diff
@@ -0,0 +1,281 @@
...
+			if (timestamp > maxDateTime) {
+				return true;
+			}
+		}
+		if (this.isDisabledDate && this.isDisabledDate(dateObj)) {
+			return true;
+		}
+	}
+
+	get displaySelectedDate() {
```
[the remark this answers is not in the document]
esmailpour-hosein: Yeah, that would be nice, but in terms of consistency, I think we should wrap it in our component to provide couple of options. We'll discuss it in the next iteration of the component based on our needs. Good idea 👍

packages/components/date-picker-overlay/README.md
```diff
@@ -0,0 +1,156 @@
...
+| pageDate          | pageDate          | String   |         |                                            |
+| selectedDate      | selectedDate      | String   |         |                                            |
+| deselectable      | deselectable      | Boolean  | false   |                                            |
+| disabledDateCheck | disabledDateCheck | Function |         |                                            |
+
+## ➤ Events
+
+| Name                 | Description |
+| -------------------- | ----------- |
+| selected-date-change |             |
```
aleks-elkin, line 50: Looks like a bug in docs generator, same event name is duplicated.

packages/components/date-picker-overlay/src/date-picker-overlay.js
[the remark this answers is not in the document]
aleks-elkin: This is strange, when I paste my code snippet and run `npm run format` I get this:
```js
  return html` <span @click="${() => !day.isDisabled && this.selectDate(day.date)}" class="${classNames}">
      ${day.date.getDate()}
  </span>`;
```

Maybe this is an additional config in your IDE?

packages/components/date-picker-overlay/src/date-picker-overlay.js
```diff
@@ -0,0 +1,198 @@
...
+			day.isSelected ? 'selected' : '',
+			day.isDisabled ? 'disabled' : ''
+		].join(' ');
+		return html`<span @click="${() => !day.isDisabled && this.selectDate(day.date)}" class="${classNames}"
+			>${day.date.getDate()}</span
+		>`;
+	}
+
+	getDayObject(dateObj) {
+		// const dayOfWeek = dateObj.getDay();
```
aleks-elkin, line 95: I agree with Sonarcube, this comment seems useless and can be removed.

packages/components/date-picker-overlay/src/date-picker-overlay.js
```diff
@@ -0,0 +1,198 @@
...
+		const timestamp = dateObj.getTime();
+		return {
+			date: dateObj,
+			isToday: this.compareDates(dateObj, new Date()),
+			isSelected: this.selectedDate && this.compareDates(new Date(dateObj), new Date(this.selectedDate)),
+			isDisabled: this.disabledDateCheck(dateObj, timestamp),
+			timestamp
+			// isWeekend: dayOfWeek === 0 || dayOfWeek === 6,
+			// isSaturday: dayOfWeek === 6,
+			// isSunday: dayOfWeek === 0
```
aleks-elkin, line 105: Do we need these comments?

aleks-elkin (commented):
Is it OK that selected date is not in perfectly round circle depending on the length of the days names? I think it is visible that height is bigger than width: 
Screenshot 2021-10-01 at 15 39 23

Screenshot 2021-10-01 at 15 40 33

And we do not support names longer that 3 letters:
Screenshot 2021-10-01 at 15 43 48

esmailpour-hosein:
For the longer day names, I can make it more flexible but based on our designs I don't think we would need it.

packages/components/date-picker/stories/date-picker.stories.js
```diff
@@ -0,0 +1,96 @@
...
+			1: 1
+		},
+		1
+	);
+	const disabled = boolean('disabled', false);
+	const readonly = boolean('readonly', false);
+	const opened = boolean('opened', false);
+	const required = boolean('required', false);
+	const deselectable = boolean('deselectable', false);
+	const typeable = boolean('typeable', true);
```
aleks-elkin, line 35: Now it is `not-typable`. Please, change it :)

packages/components/date-picker/stories/date-picker.stories.js
```diff
@@ -0,0 +1,96 @@
...
+	return html`
+		<div style="max-width: 500px;">
+			<ts-date-picker
+				id="datepickerId"
+				?readonly="${readonly}"
+				?disabled="${disabled}"
+				?opened="${opened}"
+				?required="${required}"
+				?deselectable="${deselectable}"
+				?typeable="${typeable}"
```
aleks-elkin, line 70: Not it is `not-typable`. Please, change it.

packages/components/date-picker/stories/date-picker.stories.js
```diff
@@ -0,0 +1,96 @@
...
+			0: 0,
+			1: 1
+		},
+		1
+	);
+	const disabled = boolean('disabled', false);
+	const readonly = boolean('readonly', false);
+	const opened = boolean('opened', false);
+	const required = boolean('required', false);
+	const deselectable = boolean('deselectable', false);
```
aleks-elkin, line 34: This no longer exists. Please, remove it

packages/components/date-picker-overlay/README.md
```diff
@@ -0,0 +1,155 @@
...
+## ➤ Properties
+
+| Property          | Attribute         | Type     | Default | Description |
+| ----------------- | ----------------- | -------- | ------- | ----------- |
+| dir               | dir               | String   | ltr     |             |
+| translations      | translations      | Object   |         |             |
+| firstDay          | firstDay          | Number   |         |             |
+| pageDate          | pageDate          | String   |         |             |
+| selectedDate      | selectedDate      | String   |         |             |
+| deselectable      | deselectable      | Boolean  | false   |             |
```
aleks-elkin, line 43: Can you regenerate the readme? It seems outdated.

aleks-elkin (approved):
Looks good! Great work!
Just a few non-blocking comments.

esmailpour-hosein:
image

packages/components/date-picker-overlay/README.md
esmailpour-hosein: fixed it by just using the emit helper method once

packages/components/date-picker-overlay/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@tradeshift/elements.date-picker-overlay",
+  "version": "0.28.0",
```
aleks-elkin, line 3: Could you please update the version to `0.29.1` to be in sync with other packages?
https://github.com/Tradeshift/elements/blob/master/packages/core/package.json#L3

And do the same for dependencies.

packages/components/date-picker/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@tradeshift/elements.date-picker",
+  "version": "0.28.0",
```
aleks-elkin, line 3: Could you please update the version to `0.29.1`? And do the same for dependencies.

packages/components/date-picker-overlay/src/date-picker-overlay.js
```diff
@@ -0,0 +1,196 @@
...
+				${this.controlButton(() => this.changeYear(-1), 90)}
+				<span class="control-value">${this.pageDate.getFullYear()}</span>
+				${this.controlButton(() => this.changeYear(1), 270)}
+			</div>
+		`;
+	}
+
+	controlButton(onClick, rotation) {
+		return html`
+			<ts-icon
```
aleks-elkin, line 43: I think you need to import `elements.icon`.

packages/components/date-picker-overlay/src/date-picker-overlay.js
```diff
@@ -0,0 +1,196 @@
...
+				</div>
+				<div class="day-picker">
+					${this.dayNames}
+					${this.previousMonthDays.map(day => this.dayCell(day))}
+					${this.days.map(day => this.dayCell(day, true))}
+					${this.nextMonthDays.map(day => this.dayCell(day))}
+				</div>
+				${!this.required ?
+				html`
+					<ts-button
```
aleks-elkin, line 181: I think you need to import `elements.button`.

aleks-elkin (approved):
Looks good!
Added a couple more comments, but they are non-blocking.

PR #384, opened 2021-11-02
[PEAA-755] Add header-actions slot to ts-board

esmailpour-hosein:
- [PEAA-755] Add header-actions slot to ts-board
- Fix the action-select trigger button not being aligned vertically
- Make the document-generator to just log in debug mode

[PEAA-755]: https://tradeshift.atlassian.net/browse/PEAA-755?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
Screenshot 2021-11-02 at 14 40 19

### DIFF
--- a/packages/components/action-select/src/action-select.css
+++ b/packages/components/action-select/src/action-select.css
@@ -6,10 +6,7 @@
 	}
 }
 
-.relative {
+.wrapper {
 	position: relative;
-}
-
-.font-size-0 {
 	font-size: 0;
 }
--- a/packages/components/action-select/src/action-select.css
+++ b/packages/components/action-select/src/action-select.css
@@ -5,3 +5,11 @@
 		cursor: pointer;
 	}
 }
+
+.relative {
+	position: relative;
+}
+
+.font-size-0 {
+	font-size: 0;
+}
--- a/scripts/helpers.js
+++ b/scripts/helpers.js
@@ -3,7 +3,9 @@ const path = require('path');
 const xa = require('xa');
 
 const compStateLogger = function (componentName, text) {
-	xa.custom(componentName.toUpperCase(), text, { titleColor: 'yellow', backgroundColor: '#212121' });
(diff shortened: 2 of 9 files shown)


packages/components/action-select/src/action-select.css
```diff
@@ -5,3 +5,11 @@
...
 	}
 }
+
+.relative {
+	position: relative;
+}
+
+.font-size-0 {
+	font-size: 0;
+}
```
aleks-elkin, line 15: Why not join these two classes in one? Both of them are applied to only one element.

scripts/helpers.js
```diff
@@ -3,7 +3,10 @@ const path = require('path');
 const xa = require('xa');
 
 const compStateLogger = function (componentName, text) {
-	xa.custom(componentName.toUpperCase(), text, { titleColor: 'yellow', backgroundColor: '#212121' });
+	const debugMode = false;
+	if (debugMode) {
+		xa.custom(componentName.toUpperCase(), text, { titleColor: 'yellow', backgroundColor: '#212121' });
+	}
```
aleks-elkin, line 9: Interesting way to comment out a line :)

aleks-elkin (approved):

packages/components/action-select/src/action-select.css
esmailpour-hosein: I deliberately did that, since I wanted to talk to you about atomic styling approach similar to Tailwind. I tried it out in my pet project and started to like it. Seems this approach can be a good fit for Elements. I think we should talk about it in one of our TechTalks.
I'll change this back to single class name for now, for sake of consistency.

scripts/helpers.js
esmailpour-hosein: It was suppose to be temporary until I replace it with an env variable :D I'll change it.

PR #389, opened 2022-01-05
feat: [PEAA-851] add an autogeneration of types for elements

aleks-elkin:
Added a script to autogenerate TypeScript d.ts files for each element.
Each component exports two interfaces:
- Interface with HTML attributes (name example: `TSAsideHTMLAttributes`) which could be useful for React jsx type check
- Interface with component properties (`TSAside`) which contains properties for typescript.

Both interfaces are excluding internal attributes and properties.

Added autorun of this script on pre-commit hook and on `preversion` stage.

### DIFF
--- a/packages/components/basic-table/types/basic-table.d.ts
+++ b/packages/components/basic-table/types/basic-table.d.ts
@@ -1,6 +1,29 @@
-export declare class TSBasicTable {
-	dir?: 'rtl' | 'ltr' | 'auto';
-	cols: any[];
-	data: any[];
-	selectedIds?: any[];
+export interface TSBasicTableHTMLAttributes {
+	/**  Direction of the component 'rtl' or 'ltr'  */
+	dir?: string;
+
+	/**  <br> List of columns configs, including: <br> - property: Property key of the column in data object, <br> - value: value of the title of column', <br> - visibility?: Which screen sizes this column should be visible -> 'always-visible'(default) or 'desktop-only' or 'mobile-only', <br> - size?: 'small' or 'medium' or 'large', <br> - display?: 'left' or 'right' or 'center', <br> - renderer?: you can pass a renderer function to customize the content of the cells in this column, args: (cellValue, rowObject) <br>  */
+	cols?: [];
+
+	/**  List of selected rows ids (caveat: the row should include id property)   */
+	selectedIds?: [];
+
+	/**  List of rows data objects  */
+	data?: [];
+
+}
+
+export interface TSBasicTable {
+	/**  Direction of the component 'rtl' or 'ltr'  */
+	dir?: string;
+
+	/**  <br> List of columns configs, including: <br> - property: Property key of the column in data object, <br> - value: value of the title of column', <br> - visibility?: Which screen sizes this column should be visible -> 'always-visible'(default) or 'desktop-only' or 'mobile-only', <br> - size?: 'small' or 'medium' or 'large', <br> - display?: 'left' or 'right' or 'center', <br> - renderer?: you can pass a renderer function to customize the content of the cells in this column, args: (cellValue, rowObject) <br>  */
+	cols?: [];
+
+	/**  List of selected rows ids (caveat: the row should include id property)   */
+	selectedIds?: [];
+
+	/**  List of rows data objects  */
+	data?: [];
+
 }
--- a/packages/components/date-picker-overlay/types/date-picker-overlay.d.ts
+++ b/packages/components/date-picker-overlay/types/date-picker-overlay.d.ts
@@ -1,0 +1,33 @@
+export interface TSDatePickerOverlayHTMLAttributes {
+	dir?: string;
+
+	translations?: object;
+
+	firstDay?: number;
+
+	pageDate?: string;
+
+	selectedDate?: string;
+
+	required?: boolean;
+
+	disabledDateCheck?: function;
+
+}
+
+export interface TSDatePickerOverlay {
+	dir?: string;
+
+	translations?: object;
+
+	firstDay?: number;
+
+	pageDate?: string;
+
+	selectedDate?: string;
+
+	required?: boolean;
+
+	disabledDateCheck?: function;
+
+}
--- a/scripts/generate-types.js
+++ b/scripts/generate-types.js
@@ -1,0 +1,83 @@
+const path = require('path');
+const { EOL } = require('os');
+const fs = require('fs').promises;
+const {
+	getComponentNames,
+	constants,
+	compStateLogger,
+	readComponentFile,
+	capitalizeFirstLetter,
+	camelize
+} = require('./helpers');
+
+const generatePropertyLine = (name, type, description) => {
+	let result = '';
+	if (description) {
+		result += `\t/** ${description} */${EOL}`;
+	}
+	if (name.includes('-')) {
+		// if name has dashes we put it in quotes
+		name = `"${name}"`;
+	}
+	// Replace 'Array' with '[]'.
+	// Better to use Array<T> but we don't have information about structure of array items now.
+	if (type.toLowerCase() === 'array') {
+		type = '[]';
+	}
+	result += `\t${name}?: ${type.toLowerCase()};${EOL}${EOL}`;
+	return result;
+};
+
+const getExposedProperties = srcProps =>
+	srcProps.filter(
+		// filter out empty properties and internal properties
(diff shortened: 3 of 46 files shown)


wejendorp:
This is awesome, very nice!

Will it work with a "default" typescript setup like this? Basically does autocomplete work without any extra effort?
Also, would it be possible to expose React bindings/types as well? :thinking:

aleks-elkin:
> Basically does autocomplete work without any extra effort?
> Also, would it be possible to expose React bindings/typ

@wejendorp Unfortunately it will require some setup, e.g. in react you will need to manually import Interface and specify that you have a ref ff this type: https://github.com/aleks-elkin/react-typescript-web-components/blob/master/src/components/ElementsButton.tsx#L24

For react jsx developers have to create something like this: https://github.com/aleks-elkin/react-typescript-web-components/blob/master/src/declarations.d.ts
But they should use HTMLAttributes there (I will update my repo after this PR will be released).

If we had support only React we can do it inside the each component and React developers will get this out of the box. 
E.g., we just add these lines for each component:
```ts
import React from "@types/react";

...

declare global {
  namespace JSX {
    interface IntrinsicElements {
       "ts-aside": TSAsideHTMLAttributes & React.DetailedHTMLProps<React.HTMLAttributes<HTMLElement>, HTMLElement>;
		}
	}
}
```
But we need to support all frameworks and, for example, it will conflict with Vue types.
So I think to create a separate package (and maybe repo) with react jsx bindings only so developers can only import it and do not bother to manually create the file mentioned above. Downside of this approach is that it will always have one specific version and real projects can use elements of different versions (but we do not advise to do so). So there could be a situation when our react typings expose wrong properties of the element.

So, probably a manual `declarations.d.ts` in each repo would be the most accurate way because it will rely on typings of actually used versions of elements.

wejendorp:
@aleks-elkin I understand. If we put the react-typings in a package in this monorepo it would be easier to keep it in sync than if we put it in a separate repo. I'm find with having to install `@tradeshift/elements.react` or something like that. :+1:

Just food for thought, because I remember those declaration files that you link to and find them really confusing :see_no_evil:

esmailpour-hosein:
Awesome job @aleks-elkin 🎉 
I'm still looking into the code but just wanted to put this comment what you guys are discussing:

> Downside of this approach is that it will always have one specific version and real projects can use elements of different versions (but we do not advise to do so). So there could be a situation when our react typings expose wrong properties of the element.

My suggestion is to create a separate directory, `react` or `react-types` under packages and config lerna.json to pick them and then in there we can automatically generate the correlated package for each component types with the correct versioning for each of them, so the component and the types can be in sync.

packages/components/basic-table/types/basic-table.d.ts
```diff
@@ -1,6 +1,29 @@
...
-	dir?: 'rtl' | 'ltr' | 'auto';
-	cols: any[];
-	data: any[];
-	selectedIds?: any[];
+export interface TSBasicTableHTMLAttributes {
+	/**  Direction of the component 'rtl' or 'ltr'  */
+	dir?: string;
+
+	/**  <br> List of columns configs, including: <br> - property: Property key of the column in data object, <br> - value: value of the title of column', <br> - visibility?: Which screen sizes this column should be visible -> 'always-visible'(default) or 'desktop-only' or 'mobile-only', <br> - size?: 'small' or 'medium' or 'large', <br> - display?: 'left' or 'right' or 'center', <br> - renderer?: you can pass a renderer function to customize the content of the cells in this column, args: (cellValue, rowObject) <br>  */
+	cols?: [];
```
esmailpour-hosein, line 6: Is there a reason that we have `cols` and `data` as optional?

packages/components/date-picker-overlay/types/date-picker-overlay.d.ts
```diff
@@ -0,0 +1,33 @@
...
+
+	pageDate?: string;
+
+	selectedDate?: string;
+
+	required?: boolean;
+
+	disabledDateCheck?: function;
+
+}
```
esmailpour-hosein, line 16: Not related to this PR but we should also think about a way to have the internal component like these in a separate section probably and exclude them from some stuff like type generation.

packages/components/basic-table/types/basic-table.d.ts
aleks-elkin: No, I just don't know how to decide is it a required or optional property. Probably we can make all attributes optional (otherwise it will mess with jsx complaining that developer have to specify all the attributes) and all properties non optional by default.

esmailpour-hosein: I think for now this is good enough to just have them as optional even though it's not true. For some properties we can know that they are not required, based on being boolean, or having default value, but not a reliable and enough. We should discuss how we should mark them in the code, specially since for some components slots are actually required and we should let developers know that.

scripts/generate-types.js
```diff
@@ -0,0 +1,72 @@
+const path = require('path');
+const { EOL } = require('os');
+const fs = require('fs').promises;
+const { getComponentNames, constants, compStateLogger } = require('./helpers');
+
+function capitalizeFirstLetter(string) {
+	return string.charAt(0).toUpperCase() + string.slice(1);
+}
+
+const camelize = s => s.replace(/-./g, x => x[1].toUpperCase());
```
esmailpour-hosein, line 10: maybe we can move these to helpers?

scripts/generate-types.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+(async function () {
+	// read the data from the src properties
+	// get the component names
+	const componentNames = getComponentNames();
+	for (const compName of componentNames) {
+		// get the component src properties
+		compStateLogger(compName, 'Reading src/properties.json ...');
+		const componentDir = path.join(process.cwd(), `${constants.componentsPath}/${compName}`);
+		const srcProps = await fs.readFile(`${componentDir}/docs/src/properties.json`, 'utf8');
```
esmailpour-hosein, line 42: it would be nice if we create some helper functions to get these as well.

scripts/generate-types.js
```diff
@@ -0,0 +1,72 @@
...
+		// Generate interface for js properties
+		typesFileContent += `export interface TS${className} {${EOL}`;
+		properties.forEach(async property => {
+			typesFileContent += generatePropertyLine(property.Property, property.Type, property.Description);
+		});
+
+		typesFileContent += `}${EOL}`;
+
+		await fs.writeFile(typesFilePath, typesFileContent);
+		compStateLogger(compName, 'Added type definitions.');
```
esmailpour-hosein, line 70: nit, I think we can make some smaller functions like `getExposedProperties`, `getHtmlAttributeInterfaces`, `getJsPropertiesInterfaces`, `getTypesFileContent`, ... to make it even cleaner.

esmailpour-hosein (approved):
Looks great! Thanks for moving this forward 👍 
Just some minor improvement suggestions.

packages/components/date-picker-overlay/types/date-picker-overlay.d.ts
aleks-elkin: The easiest solution would be just to have a list of these components. Another solution is to add a JSDoc comment on top of the class and process it in our doc generator so we can filter later by this comment.
Anyway, I think we don't have that many internal components and I don't think that we will move some components between internal and public states very often, so we can go with simple list.

scripts/generate-types.js
esmailpour-hosein: Yeah, there's a tradeoff, I think I'm still high on uncle Bob clean code talk :D

esmailpour-hosein: but in general I kinda like to get a general understanding of what is happening without the detail in the first look and if I cared I can find what is happening underneath. Although it would make it easier to test and maintain it in long term. But I understand how during the implementation it can feel too much, since you have to jump all over the place 👍

aleks-elkin: In this particular case i think it is a grey area, because the function is big, so it is ok to have a feeling to split it in smaller ones. But this is also a script, not an actual function that is used in the project, so it is kinda standalone, so I think there is no clear answer, what is the right way to do in such cases :)

esmailpour-hosein: Yeah, agree! :)

tradeshift-larsbojensen: Agree, especially if these functions can be expected to be used only here 👍

PR #401, opened 2022-01-19
[PEAA-842] Action select positioning issue on top-right corner in relative mode

tradeshift-larsbojensen:
As documented in https://tradeshift.atlassian.net/browse/PEAA-842 , the ts-select-action dropout will in some cases miscalculate if it fits in on the right side of its anchor. The basic issue here is that the component tries to calculate that based on, among other factors, the width of the rendered dropout. So it depends on content, and even on wrapping that content. It will re-calculate on window resize, getting valid values.
The changes will reduce this wrong behaviour, but it is probably not bulletproof.
Another issue is that it would sometimes pull dropouts partly off-viewport (opening to the left on narrow enough screens). The changes fix that. The issue:

old-issue-opens-bottom-left
old-issue-opens-top-left

The issue referenced in the task:

image-20211217-131633
image-20211217-131642
.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ loadPolyfills().then(() => import('./my-app.js'));
 
 - Installation
 
-```hell
+```shell
 $ npm i @webcomponents/webcomponentsjs --save
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -134,7 +134,7 @@ loadPolyfills().then(() => import('./my-app.js'));
 
 - Installation
 
-```hell
+```shell
```
esmailpour-hosein, line 137: :)))

esmailpour-hosein (approved):
All looks good, but can you please provide a bit more information in the PR description?

aleks-elkin (approved):

PR #408, opened 2022-01-24
Add missing types for new components and fix new component template to add them by default

aleks-elkin:
Our new components did not contain type definitions references in their package.json and these typings did not published.
This Pr fixes this issue by adding missing types definitions and adjusting new component template to have them by default.


esmailpour-hosein (approved):
👍

PR #415, opened 2022-02-08
PEAA-593: add search suggestions

aleks-elkin:
Added an optional list of suggestions to a search component.
Also, added a loading state to a `select-menu` component to show loading spinner when there is an asynchronous loading of the data.

### Screencast:
https://user-images.githubusercontent.com/55530374/153043522-ac9fdbd1-fa13-4bdd-870c-95e53cc190ca.mov

Moreover:
- changed a dev setup to use `@web/dev-server` and ES modules without babel transpilation for local dev environment. I think it will improve the time of recompilation after each change.
- added a list of components pages to a root index.html file. The idea is to use it as a list of references to each component, so we could develop and test each component on a separate page.


packages/components/search/src/search.js
```diff
@@ -85,14 +145,36 @@ export class TSSearch extends TSElement {
...
 				if (this.focused) {
 					// set focus only when component is defined.
-					window.customElements.whenDefined(ELEMENT_TAG).then(() => this.shadowRoot.getElementById('input').focus());
+					window.customElements.whenDefined(ELEMENT_TAG).then(() => {
+						this.shadowRoot.getElementById('input').focus();
+						if (this.showSuggestions && this.focused) {
+							this.opened = true;
+						} else {
+							this.opened = null;
+						}
```
tradeshift-larsbojensen, line 154: Not boolean? Does it need a third state (the null)? If boolean, it could be:

this.opened = this.showSuggestions && this.focused

tradeshift-larsbojensen (commented):
I would like to see this added to the storybook.

packages/components/select-menu/src/select-menu.js
```diff
@@ -132,29 +142,37 @@ export class TSSelectMenu extends TSElement {
 		);
 		const showApplyButtonContainer = !this.noApplyButton && this.multiselect && this.dirty;
 		return html`<div id="listContainer">
+			${this.loading
+				? html`<div style="padding-top: 10px;">
```
esmailpour-hosein, line 146: Should we do it in css file?

packages/components/search/src/search.js
```diff
@@ -119,24 +202,53 @@ export class TSSearch extends TSElement {
...
-					?autofocus="${this.autofocus}"
-					.value="${this.value}"
-					.placeholder="${this.placeholder}"
-					@keydown="${this.handleKeyDown}"
-					@input="${this.handleInput}"
-					@focus="${this.handleFocus}"
-					@blur="${this.handleBlur}"
-				/>
-				<div class="icon-container">${this.showCancel ? this.cancelIcon : ''}</div>
+				${this.opened && this.showSuggestions ? this.suggestionsOverlay : ''}
```
esmailpour-hosein, line 251: I prefer to use parentheses in these kinda cases to be more clear.

esmailpour-hosein (approved):
Great job! 🚀 
Just some small comments.

packages/components/select-menu/src/select-menu.js
esmailpour-hosein: And also, maybe increase it to --ts-unit?

esmailpour-hosein:
I have some mixed feelings about the  name `suggestion` for the search results. I was thinking maybe we can use aliases for this property to cover both cases from the exposed API point of view, what do you think?

aleks-elkin:
> I have some mixed feelings about the name `suggestion` when it used for the search results. I was thinking maybe we can use aliases for this property to cover both cases (recent searched items/suggestions and actual search results) from the exposed API point of view, what do you think?

@esmailpour-hosein yes, I agree that we need to be more clear with this property naming. I'm.not sure about how to implement aliases without introducing an additional and probably complex logic, so I'm thinking about to name it with something more generic, like `select-items` or `dropdown-items`. What do you think?

packages/components/search/src/search.js
```diff
@@ -23,7 +29,17 @@ export class TSSearch extends TSElement {
 			/** Message when an input is empty */
 			placeholder: { type: String, reflect: true },
 			/** The current value */
-			value: { type: String, reflect: true }
+			value: { type: String, reflect: true },
+			/** Translated messages for the user locale */
+			translations: { type: Object, reflect: true },
```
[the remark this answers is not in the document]
aleks-elkin: So, I added this type definition:
```js
 @type {{ loading: string, no_items: string }}
```
 Not much, but at least something to start with.

PR #416, opened 2022-02-14
Update the event comment in search component and fix error in select-like components

aleks-elkin:
Added two fixes:
- updated the event comment in search component to reflect a new logic, that it now triggered when user selects an item from dropdown
- delete a closeOverlay reference after it is called because overlay already closed and it will throw an error on next call. It could happen in this case: 
  - open select component (it will keep the reference to the function)
  -  when it got attached to DOM and detached several times (e.g. in aside or modal) it will try to call this function again. But the overlay no longer exists (it got destroyed after first call).
If we remove this reference after a call it should solve the issue.

### DIFF
--- a/packages/components/search/src/search.js
+++ b/packages/components/search/src/search.js
@@ -138,7 +138,7 @@ export class TSSearch extends TSElement {
 					return;
 				}
 				/**
-				 * Emitted on every user's change in a search input
+				 * Emitted on every user's change in a search input or when user selects an item from the provided `dropdownItems`
 				 * @payload search input value
 				 */
 				this.dispatchCustomEvent('change', newVal);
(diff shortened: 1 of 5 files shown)


packages/components/search/src/search.js
```diff
@@ -138,7 +138,7 @@ export class TSSearch extends TSElement {
 					return;
 				}
 				/**
-				 * Emitted on every user's change in a search input
+				 * Emitted on every user's change in a search input or when user selects from a dropdown
```
esmailpour-hosein, line 141: I think it would be more clear if we say this instead: 
```
				 * Emitted on every user's change in a search input or when user selects an item from the provided `dropdownItems`
```
What do you think?

esmailpour-hosein (approved):

packages/components/search/src/search.js
aleks-elkin: Sounds good to me. I will update it.

esmailpour-hosein (approved):

PR #418, opened 2022-02-16
[PEAA-169] Add tag element

tradeshift-larsbojensen:
I didn't add storybook nor a happo test. I want to discuss and possibly merge the component first

### DIFF
--- a/packages/components/tag/src/tag.css
+++ b/packages/components/tag/src/tag.css
@@ -1,0 +1,238 @@
+:host {
+	min-width: 1em;
+	margin: 0 var(--ts-unit-half) var(--ts-unit-half) 0;
+	display: inline-flex;
+	max-width: 100%;
+}
+
+/* Container base styles ***************************/
+
+:host .container {
+	display: inline-flex;
+	align-items: center;
+	margin: 0;
+	max-width: 100%;
+	min-height: var(--ts-unit-plus);
+	padding: 0 var(--ts-unit-half);
+	background-color: var(--ts-color-gray-lighter);
+	color: var(--ts-color-black);
+	border-radius: var(--ts-radius);
+	font-size: var(--ts-fontsize);
+	font-family: var(--ts-fontstack);
+	font-style: normal;
+}
+
+:host([clickable]) .container:hover {
+	background-color: var(--ts-color-gray-light);
+}
+
+/* Focused tag *******************************************/
+
+:host(:focus-within) .container {
+	outline: none;
+}
+:host([type='success']:focus-within) {
+	outline-color: var(--ts-color-green-dark);
+}
+:host([type='warning']:focus-within) {
+	outline-color: var(--ts-color-orange-dark);
+}
+:host([type='warning-lite']:focus-within) {
+	outline-color: var(--ts-color-orange-lighter);
+}
+:host([type='danger']:focus-within) {
+	outline-color: var(--ts-color-red-dark);
+}
+:host([type='blue']:focus-within) {
+	outline-color: var(--ts-color-blue-dark);
+}
+:host([type='blue-lite']:focus-within) {
+	outline-color: var(--ts-color-blue-light);
+}
+:host(:not([type]):focus-within) {
+	outline-color: var(--ts-color-gray-light);
+}
+
+/* Labels and values (text content) .................................*/
+
+.label {
+	text-transform: uppercase;
+	font-weight: var(--ts-fontweight-semibold);
+}
+
+.label,
+.value {
+	white-space: nowrap;
+	overflow: hidden;
+	text-overflow: ellipsis;
+}
+
+.value {
+	flex: 1;
+}
+
+/* Colors ....................................................*/
+
+:host([type='success']),
+:host([type='warning']),
+:host([type='danger']),
+:host([type='blue']) {
+	& > .container {
+		color: var(--ts-color-white);
+	}
+}
+
+:host(:not([busy]):not([locked])[type='success']) {
+	& .container {
+		background-color: var(--ts-color-green);
+	}
+}
+
+:host(:not([busy]):not([locked])[type='success'][clickable]) {
+	& .container:hover {
+		background-color: var(--ts-color-green-dark);
+	}
+}
+
+:host(:not([busy]):not([locked])[type='warning']) {
+	& .container {
+		background-color: var(--ts-color-orange);
+	}
+}
+
+:host(:not([busy]):not([locked])[type='warning'][clickable]) {
+	& .container:hover {
+		background-color: var(--ts-color-orange-dark);
+	}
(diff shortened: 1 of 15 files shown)


packages/components/tag/src/utils/constants.js
```diff
@@ -0,0 +1,18 @@
...
+
+export const labels = {
+	SINGLE_LABEL: ['Label 1'],
+	MULTIPLE_LABELS: ['Label 1', 'Label 2']
+};
+
+export const values = {
+	SINGLE_VALUE: ['value 1'],
+	MULTIPLE_VALUES: ['value 1', 'value 2']
+};
```
esmailpour-hosein, line 18: I think we should not use the utils for the stories-only usage and everything related to the stories should live inside that directory. In this case I would put it directly to the happo file.

packages/components/tag/src/tag.js
```diff
@@ -0,0 +1,93 @@
...
+			/** Text direction: 'rtl' or 'ltr' */
+			dir: { type: String, reflect: true },
+			/** Type: success, warning, warning-lite, danger, blue, blue-lite */
+			type: { type: String, reflect: true },
+			/** Can the item be deleted? */
+			deletable: { type: Boolean, reflect: true },
+			/** Should the tag look like clicking it triggers an action? */
+			clickable: { type: Boolean, reflect: true },
+			/** Is the item locked? */
+			locked: { type: Boolean, reflect: true },
```
esmailpour-hosein, line 21: Just a general suggestion, I think we should also try to mention what would be reflected in the UI or behaviour by changing these kinda values.

packages/components/tag/src/tag.js
```diff
@@ -0,0 +1,93 @@
...
+		/** constraint event; do not trigger events up the DOM tree */
+		event.stopPropagation();
+		if (!this.busy) {
+			this.dispatchCustomEvent('delete-tag', event);
+		}
+	}
+
+	handleClickEvent(event) {
+		if (!this.busy) {
+			this.dispatchCustomEvent('click-tag', event);
```
esmailpour-hosein, line 45: Please add a comment, so our doc generator would pick it up for our readme file, [see here](https://github.com/Tradeshift/elements/blob/master/packages/components/search/src/search.js#L140)

packages/components/tag/src/tag.js
```diff
@@ -0,0 +1,93 @@
...
+
+	get direction() {
+		return this.dir || this.bodyDir;
+	}
+
+	handleDeleteEvent(event) {
+		/** constraint event; do not trigger events up the DOM tree */
+		event.stopPropagation();
+		if (!this.busy) {
+			this.dispatchCustomEvent('delete-tag', event);
```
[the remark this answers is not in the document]
esmailpour-hosein: I think we can also send some of tag information like key and value, so developers can have one listener and distinguish items without looking into target detail, if needed.

packages/components/tag/src/tag.js
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		if (this.locked) {
+			return html`
+				<span class="icon">
+					<ts-icon icon="lock" size="large"></ts-icon>
+				</span>
+			`;
+		}
+
```
esmailpour-hosein, line 74: by swapping their places, we can get rid of the `!this.locked` condition on the first `if`

packages/components/tag/src/tag.js
```diff
@@ -0,0 +1,93 @@
...
+		return html`<div
+			dir=${this.direction}
+			class=${this.classes}
+			@click=${this.handleClickEvent}
+			@keyup=${this.handleClickEvent}
+		>
+			${this.labels &&
+			html`<span class="label">${this.labels.join(', ')}${this._hasBothLabelsAndValues() ? `:` : null}</span>`}
+			${this._hasBothLabelsAndValues() ? html`<span>&nbsp;</span>` : null}
+			${this.values && html`<span class="value">${this.values.join(', ')}</span>`} ${this.icons}
```
esmailpour-hosein, line 88: We should refactor this, it's not easy to read and follow. 
Maybe we can have multiple getters like `icons` to separate these from each other and simpler render block

esmailpour-hosein: we can do it in another PR, not a blocker

tradeshift-larsbojensen: Yeah, I was thinking the same, but decided they were not too complicated to read, also considering the simplicity of the entire component markup. I agree it's a nice idea, though.

packages/components/tag/src/tag.js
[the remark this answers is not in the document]
tradeshift-larsbojensen: Checking keys and/or values is looking into detail, just not of the target.detail. I suggest developers add id's to ts-tags in order to identify them. But let's discuss this, maybe you are right.

packages/components/tag/src/utils/constants.js
tradeshift-larsbojensen: I agree, I just thought it was a decided standard. I also think it's over the top in a case like this.

packages/components/tag/src/tag.css
```diff
@@ -0,0 +1,238 @@
...
+}
+
+/* Container base styles ***************************/
+
+:host .container {
+	display: inline-flex;
+	align-items: center;
+	margin: 0;
+	max-width: 100%;
+	min-height: 30px;
```
esmailpour-hosein, line 15: please use --ts-unit-plus

packages/components/tag/src/tag.css
```diff
@@ -0,0 +1,238 @@
...
+
+:host(:not([type])[clickable]) .container {
+	color: var(--ts-color-blue);
+}
+
+/* Icons ........................................................*/
+
+:host([locked]),
+:host([deletable]) {
+	& .container[dir='ltr'] {
```
esmailpour-hosein, line 179: nit, Generally we have default state, which would be ltr and add specific dir='rtl' overrides.

packages/components/tag/src/tag.css
```diff
@@ -0,0 +1,238 @@
...
+		padding-left: 0;
+	}
+}
+
+.icon {
+	display: inline-flex;
+	align-items: inherit;
+	justify-content: center;
+	width: 30px;
+	height: 30px;
```
esmailpour-hosein, line 195: please use --ts-unit-plus

esmailpour-hosein (approved):
Great job! 🎉 
Nothing big, just some suggestions and minor improvements.

packages/components/tag/src/tag.js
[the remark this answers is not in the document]
esmailpour-hosein: Let's keep it as it is for now.

PR #420, opened 2022-02-23
[PEAA-592] Select: add custom filtering + minor bug fixes 

esmailpour-hosein:
- [feat(select): add case-sensitive filtering to select component](https://github.com/Tradeshift/elements/pull/420/commits/ee8182f4ca559e2815d6ee4e54f638e20b528dec) 
   `caseSensitive` property or `case-sensitive` can be passed to the select and select-menu components to make the filtering/search case-sensitive.
- [feat(select): add custom filtering to select component](https://github.com/Tradeshift/elements/pull/420/commits/690abf8ff1cd4164dcfead0d33e1d49c0f934cab) 
    Now it's possible to do custom filtering beside the basic filter that is done by default for item.title text. Now developers can listen to `filter-value-change` event, to get the user inputed value for search/filter and pass `filteredItems` to override the default internal filter/search through the already provided items. To make it possible to do the filtering asynchronously, added loading state to the select component.

    Note: Items that are provided for `filteredItems`, should be also included in `items`, which means developers should make sure they add them to items list if they are not already there and not remove other items for multi select, since they can be already in current selection state.
    
    
 - [fix(select): show the View selection button in multi select with no apply](https://github.com/Tradeshift/elements/pull/420/commits/c70c4f606e9b0548235e6ab539ed124acaa6fce3) 
    The `View selection` button was getting hidden by setting the noApplyButton property, which was not the correct behaviour. The `View selection` button should also be visible when there is already some selected items for the select, even if the user did not change the selection.
    
- [fix(select): do not set input value for multi select and no apply](https://github.com/Tradeshift/elements/pull/420/commits/b5abd2129c38e464a719f08b6b7bdaa8599635e6) 
    In the no apply button mode, since we change the selection without changing the `opened` state of the component, it would set the input value, which was overriding the entered filter value and causing bad user experience. Since there's no scenario that we want to set the input value while select-menu is open, just ignoring the update input value request.
    
### new added commits:

- [fix(select): peaa-941 current selection clears on search input](https://github.com/Tradeshift/elements/pull/420/commits/15b80caaa96ace990e9616a317492c5d74397258) 
   After user started typing in the search input of the select component, the currentSelection was overwritten in select-menu.js by the selected items of the select.js.
Now it updates the current selection at the beginning and when it updates, so the currentSelection would remain only for the staging state of multiselect. 

- [chore(select): add more stories and test](https://github.com/Tradeshift/elements/pull/420/commits/095095d21767713f6ffed39fd3c1c34d55b98efc) 
    - Added stories for custom filtering and no apply button mode: added stories with more information about the component behaviour and what can be done with it.
   - Fixed missing description for events and props

### DIFF
--- a/packages/components/select-menu/src/select-menu.js
+++ b/packages/components/select-menu/src/select-menu.js
@@ -1,4 +1,4 @@
-import { TSElement, unsafeCSS, html, customElementDefineHelper } from '@tradeshift/elements';
+import { customElementDefineHelper, html, TSElement, unsafeCSS } from '@tradeshift/elements';
 import '@tradeshift/elements.button';
 import '@tradeshift/elements.button-group';
 import '@tradeshift/elements.icon';
@@ -20,6 +20,8 @@ export class TSSelectMenu extends TSElement {
 			disabled: { type: Boolean, reflect: true },
 			/** List of available options. Item must have 'id' and 'title', it can also have an 'icon' which is the name of the icon */
 			items: { type: Array, reflect: true },
+			/** List of filtered options based on the select filter input value. You should update the  */
+			filteredItems: { type: Array, reflect: true, attribute: 'filtered-items' },
 			/** Allow users to select several options or not. */
 			multiselect: { type: Boolean, reflect: true },
 			/** Do not show the apply button and directly emit select-menu-changed when the selection changes.
@@ -140,11 +142,16 @@ export class TSSelectMenu extends TSElement {
 	}
 
 	get displayedItems() {
+		if (this.showSelectedOnly) {
+			return this.items.filter(item => this.currentSelection.indexOf(item.id) > -1);
+		}
+		if (this.filteredItems && this.filterValue) {
+			return this.filteredItems;
+		}
 		const searchString = this.caseSensitive ? this.filterValue : this.filterValue?.toLowerCase();
 		return this.items.filter(item => {
 			const itemTitle = this.caseSensitive ? item.title : item.title.toLowerCase();
-			const isSelected = this.currentSelection.indexOf(item.id) > -1;
-			return itemTitle.indexOf(searchString) > -1 && (this.showSelectedOnly ? isSelected : true);
+			return itemTitle.indexOf(searchString) > -1;
 		});
 	}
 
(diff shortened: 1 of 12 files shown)


packages/components/select-menu/src/select-menu.js
```diff
@@ -20,6 +20,8 @@ export class TSSelectMenu extends TSElement {
 			disabled: { type: Boolean, reflect: true },
 			/** List of available options. Item must have 'id' and 'title', it can also have an 'icon' which is the name of the icon */
 			items: { type: Array, reflect: true },
+			/** List of filtered options based on the select filter input value. You should update the  */
```
aleks-elkin, line 23: I think you lost the part of sentence about what they should update.

packages/components/select-menu/src/select-menu.js
```diff
@@ -149,8 +171,8 @@ export class TSSelectMenu extends TSElement {
 				: html`
 				<ul>
 					${
-						filteredItems.length > 0
-							? filteredItems.map(
+						this.displayedItems.length > 0
+							? this.displayedItems.map(
```
aleks-elkin, line 175: You call getter of `displayedItems` twice, but it has a complex logic inside. Consider to call it once and store the value in a local variable.

packages/components/select-menu/src/select-menu.js
```diff
@@ -122,25 +129,40 @@ export class TSSelectMenu extends TSElement {
...
+	}
+
 	/** @private */
 	get selectButton() {
 		return html`
-			<ts-button-group>
+			<ts-button-group class="apply-button-container ${this.showApplyButtonContainer ? 'show' : 'hide'}">
 				<ts-button type="primary" @click=${this.applySelection}
 					>${this.translations.select} ${this.currentSelection.length}</ts-button
 				>
```
aleks-elkin, line 142: Strange line break in a closing tag (I assume it is from prettier). Please, fix.

packages/components/select/src/select.js
```diff
@@ -20,6 +20,12 @@ export class TSSelect extends TSElement {
 		this._translations = Object.assign({}, translations);
 		this.handleInputDebounced = helpers.debounceEvent(() => {
 			this.filterValue = this.inputValue;
+			/**
+			 * Emitted when filter value of the select changes. You can listen to this to update the
```
aleks-elkin, line 24: I think you lost end of the sentence. Please, add the information, what they can update from this event.

aleks-elkin (commented):
Looks good, I have just a few minor comments.

aleks-elkin (approved):

PR #423, opened 2022-02-28
fix(tag): fixes padding issue in locked and deletable items; also cle…

tradeshift-larsbojensen:
…ans up code a bit

https://tradeshift.atlassian.net/browse/PEAA-950

Uses a logical CSS selector for removing padding after an icon (locked and deletable tags have an icon at the end of the tag). Before the change, locked and deletable tags would have no padding at the right side of the tag in RTL mode; i.e. where the tag text/content begins. The padding-inline-end saves lines/rules compared to a solution controlling padding-right and padding-left.
The commit also cleans up a typo, removes unnecessary CSS rule nestings, and makes ts-tag element render code easier to read.

### DIFF
--- a/packages/components/tag/src/tag.css
+++ b/packages/components/tag/src/tag.css
@@ -49,7 +49,8 @@
 :host([type='blue-lite']:focus-within) {
 	outline-color: var(--ts-color-blue-light);
 }
-:host(:not([type]):focus-within) {
+:host(:not([type]):focus-within),
+:host([locked]:focus-within) {
 	outline-color: var(--ts-color-gray-light);
 }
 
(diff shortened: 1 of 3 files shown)


packages/components/tag/src/tag.js
```diff
@@ -83,17 +83,32 @@ export class TSTag extends TSElement {
...
 			dir=${this.direction}
 			class=${this.classes}
 			@click=${this.handleClickEvent}
 			@keyup=${this.handleClickEvent}
 		>
-			${this.labels &&
-			html`<span class="label">${this.labels.join(', ')}${this._hasBothLabelsAndValues() ? `:` : null}</span>`}
-			${this._hasBothLabelsAndValues() ? html`<span>&nbsp;</span>` : null}
-			${this.values && html`<span class="value">${this.values.join(', ')}</span>`} ${this.icons}
+			${this._labels} ${this._separator} ${this._values} ${this.icons}
```
tradeshift-larsbojensen, line 111: Our commit hook fixup machinery wants this in 1 line. I would have preferred this for readability:

```
			${this._labels}
			${this._separator}
			${this._values}
			${this.icons}
```

packages/components/tag/src/tag.js
```diff
@@ -83,17 +83,32 @@ export class TSTag extends TSElement {
...
 			dir=${this.direction}
 			class=${this.classes}
 			@click=${this.handleClickEvent}
 			@keyup=${this.handleClickEvent}
 		>
-			${this.labels &&
-			html`<span class="label">${this.labels.join(', ')}${this._hasBothLabelsAndValues() ? `:` : null}</span>`}
-			${this._hasBothLabelsAndValues() ? html`<span>&nbsp;</span>` : null}
-			${this.values && html`<span class="value">${this.values.join(', ')}</span>`} ${this.icons}
+			${this._labels} ${this._separator} ${this._values} ${this._icons}
```
tradeshift-larsbojensen, line 111: Our commit hook fixup makes this inline. I would prefer it like this for readability:

```
			${this._labels}
			${this._separator}
			${this._values}
			${this._icons}
```

packages/components/tag/src/tag.css
```diff
@@ -82,80 +82,56 @@
 	}
 }
 
-:host(:not([busy]):not([locked])[type='success']) {
-	& .container {
-		background-color: var(--ts-color-green);
-	}
+:host(:not([busy]):not([locked])[type='success']) .container {
```
esmailpour-hosein, line 85: Can we simplify the selectors? I think we can just override the busy and locke state if needed.

packages/components/tag/src/tag.css
```diff
@@ -177,13 +153,7 @@
...
 :host([deletable]) {
 	& .container {
-		padding-right: 0;
-	}
-}
-:host([locked]),
-:host([deletable]) {
-	& .container[dir='rtl'] {
-		padding-left: 0;
+		padding-inline-end: 0;
```
esmailpour-hosein, line 156: super cool! Didn't know about these new properties. 👍 Thanks god we don't have support IE anymore :))

packages/components/tag/src/tag.js
esmailpour-hosein: Since there's not much going on in the template anymore, I suggest we use // prettier-ignore on top of the return here.

esmailpour-hosein (approved):

packages/components/tag/src/tag.css
tradeshift-larsbojensen: It's new to me too. They are awesome, and surprisingly well supported.

esmailpour-hosein (approved):

PR #424, opened 2022-03-01
Release/v0.35.1

tradeshift-larsbojensen:
Releasing the fixes and refactoring in https://tradeshift.atlassian.net/browse/PEAA-950

### DIFF
--- a/packages/core/src/utils/busyButton.css
+++ b/packages/core/src/utils/busyButton.css
@@ -1,0 +1,34 @@
+/* Busy status. Adds an animated diagonal stripe fill to indicate loading ............*/
+
+:host([busy]) > :first-child {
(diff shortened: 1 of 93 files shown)


packages/core/src/utils/busyButton.css
```diff
@@ -0,0 +1,34 @@
+/* Busy status. Adds an animated diagonal stripe fill to indicate loading ............*/
```
aleks-elkin, line 1: I suggest to include it in a separate PR, this is not a result of release script.
Also, I don't see how this file is used anyway.

tradeshift-larsbojensen: I added the file by mistake, I will delete it

PR #426, opened 2022-03-01
chore: updates package-lock

tradeshift-larsbojensen:
I have deleted tags above 0.34.0, because the release build failed. For the same reason, the release commit for 0.35.0 has been reverted. This is to enable releasing a new version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5766,14 +5766,12 @@
             "@babel/runtime": "^7.3.1",
             "highlight.js": "~9.18.2",
             "lowlight": "~1.11.0",
-            "prismjs": "^1.8.4",
+            "prismjs": ">=1.25.0",
             "refractor": "^2.4.1"
           },
           "dependencies": {
             "prismjs": {
-              "version": "1.27.0",
-              "resolved": "https://registry.npmjs.org/prismjs/-/prismjs-1.27.0.tgz",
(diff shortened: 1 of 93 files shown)


package-lock.json
```diff
@@ -7163,305 +7163,305 @@
...
+        "@tradeshift/elements": "^0.35.0",
+        "@tradeshift/elements.icon": "^0.35.0",
+        "@tradeshift/elements.tab": "^0.35.0",
+        "@tradeshift/elements.typography": "^0.35.0"
       }
     },
     "@tradeshift/elements.tag": {
       "version": "file:packages/components/tag",
       "requires": {
         "@tradeshift/elements": "^0.33.8",
```
aleks-elkin, line 7463: Hmm, why this dependency is not updated?

tradeshift-larsbojensen: Super catch. I haven't analysed the version creating script, but I guess because it was never released.

aleks-elkin: I think it should be fixed here: https://github.com/Tradeshift/elements/blob/master/packages/components/tag/package.json

tradeshift-larsbojensen: Should I just manually make it 0.35.0?

esmailpour-hosein (approved):

PR #428, opened 2022-03-02
feat(button-group): peaa-885 button-group support horizontal/inline mode

esmailpour-hosein:
Now it's possible to use the button-group in horizontal/inline mode with `inline`
property/attibute, which would make the buttons in the group aligned, margined
and wrapped properly.

### DIFF
--- a/packages/components/button-group/src/button-group.css
+++ b/packages/components/button-group/src/button-group.css
@@ -4,3 +4,11 @@
 	display: block;
 	width: 100%;
 }
+
+:host([inline]) {
+	font-size: 0;
+}
+
+:host([inline]) ::slotted(ts-button) {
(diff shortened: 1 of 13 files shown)


packages/components/button-group/src/button-group.css
```diff
@@ -4,3 +4,16 @@
 	display: block;
 	width: 100%;
 }
+
+:host([inline]) {
+	font-size: 0;
+}
+
+:host([inline]) ::slotted(ts-button) {
+	margin-right: var(--ts-unit-half);
```
aleks-elkin, line 13: Did you try `margin-inline-end: var(--ts-unit-half);`? The result looks the same for me, and it allows to drop the css rule below ( the one for rtl case).

aleks-elkin (approved):
Great work!

PR #429, opened 2022-03-02
feat(dialog): adds notification options to dialog element

tradeshift-larsbojensen:
Task: https://tradeshift.atlassian.net/browse/PEAA-395

These changes add notification features to the dialog element:

A dialog can have a `notification` attribute and will have no cancel button, and if `no-buttons`, no buttons at all.

A `success` type `notification` will auto-close after 1500 ms, if not non-dismissable.

A `non-dismissable` notification or dialog cannot be closed except by clicking a button provided by the dialog/notification.

The changes also include a fix that makes auto-focusing a button ('accept' or 'cancel') work, although only in dialog (not notification) mode, and a new attribute is added to ts-modal, making closing on clicking the cover optional.

### DIFF
--- a/packages/components/dialog/src/dialog.js
+++ b/packages/components/dialog/src/dialog.js
@@ -1,4 +1,11 @@
-import { TSElement, unsafeCSS, html, customElementDefineHelper, validateSlottedNodes } from '@tradeshift/elements';
+import {
+	TSElement,
+	unsafeCSS,
+	html,
+	customElementDefineHelper,
+	validateSlottedNodes,
+	constants
+} from '@tradeshift/elements';
 import '@tradeshift/elements.button';
 import '@tradeshift/elements.button-group';
 import '@tradeshift/elements.icon';
@@ -14,6 +21,8 @@ export class TSDialog extends TSElement {
 		this.type = dialogTypes.CONFIRM;
 		this.focused = 'cancel';
 		this._translations = Object.assign({}, translations);
+		this.notification = false;
+		this.nonDismissable = false;
 	}
 
 	static get styles() {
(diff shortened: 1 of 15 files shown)


esmailpour-hosein (changes requested):

packages/components/dialog/src/dialog.js
```diff
@@ -22,20 +31,26 @@ export class TSDialog extends TSElement {
...
 			/** can be used for customizing the buttons text and translations */
 			translations: { type: Object, reflect: true },
 			/** set the default focus on the button, either `accept` or `cancel` */
 			focused: { type: String, reflect: true },
 			/** either `accept` or `cancel` can be used to change the button type, based on the dialog type, by default both are secondary */
 			primary: { type: String, reflect: true },
+			/** If it is a notification, no cancel button will be rendered. Notifications of type 'success' will auto-close on timeout, if they are not `non-dismissable` */
+			notification: { type: Boolean, reflect: true },
+			/** Render no buttons. This only affects notifications of type 'success' */
+			noButtons: { type: Boolean, reflect: true, attribute: 'no-buttons' },
```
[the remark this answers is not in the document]
tradeshift-larsbojensen: I think that is an excellent suggestion! The noButtons is a notifications-only, even only for success type, prop, and you should always try to avoid that. Just easily happens when when merging two components into one.

packages/core/src/utils/constants.js
```diff
@@ -33,7 +33,8 @@ export const colors = {
 export const delay = {
 	NOW: 100,
 	FAST: 200,
-	SLOW: 600
+	SLOW: 600,
+	CLOSE: 1500
```
esmailpour-hosein, line 37: I think it would be better if we move this into the component itself.  We probably should get rid of this delay object all together soon, I don't think it brings that much of value to share these values between components.

tradeshift-larsbojensen: I was thinking it makes a lot of sense to have shared timing constants, but also noticed that most (all?) components just have hardcoded values for timeouts. The `delay` object is otherwise only used by the core, in which it is located, so I think we can safely keep the object. I will make my timeout use a local value. That is in alignment with other elements. If we ever want to make that shared constants, we should decide that when planning future development of the entire library.

packages/components/dialog/src/dialog.js
```diff
@@ -98,7 +91,7 @@ export class TSDialog extends TSElement {
 			if (this.notification === true && this.type === dialogTypes.SUCCESS && !this.nonDismissable) {
 				return window.setTimeout(() => {
 					this.dismissModal();
-				}, constants.delay.CLOSE);
+				}, 1500);
```
esmailpour-hosein, line 94: 🙈 Sorry, but can you please keep it as a constant but just in the component's utils/constants ?

tradeshift-larsbojensen: on it

esmailpour-hosein (approved):

PR #436, opened 2022-03-09
fix(button): fixes missing/broken focus state styling

tradeshift-larsbojensen:
Motivation: ts-buttons are lacking a border indicating focused state. In case type=secondary it makes a focused button look like plain text.

Note that letter-spacing is removed on the buttons, because there is no custom letter-spacing in the design. It improves readability IMO, but triggers a lot of happo snapshot changes.

Design: https://www.figma.com/file/v4lzSShpqzYtQ5aQRcKAmV/UI-Library---Clean-(New-system-without-documentation)?node-id=3453%3A1

### DIFF
--- a/index.html
+++ b/index.html
@@ -26,6 +26,9 @@
 					<a href="/packages/components/board/index.html">Board</a>
 				</li>
 				<li>
+					<a href="/packages/components/button/index.html">Button</a>
+				</li>
+				<li>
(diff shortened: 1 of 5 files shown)


esmailpour-hosein:
Just for sake of reference, so we don't get a broken link or changed state later on, here is the design for it:
Screenshot 2022-03-09 at 11 52 21

esmailpour-hosein (approved):

PR #456, opened 2022-03-21
chore(deps): disable dependency-auto-update.yml cron job

esmailpour-hosein:
Since we are using Happo for visual regression testing on each PR update,
we are using our snapshots quota on these PRs, which are pretty limited.
I'm disabling it, until we either increase the quota or replace Happo
with another solution.


adinaiordache2017 (approved):
👍

aleks-elkin (approved):

PR #461, opened 2022-03-30
PEAA-879: add all icons

aleks-elkin:
Added all icons available in our design system:
Icons

Also, ran a SVGO optimizer on these icons to reduce the file size.


esmailpour-hosein (commented):
Great job! 👍 
Based on the Happo snapshots, I can see at least the arrow-up and remove are different from the old one. 
My suggestions:  
- In a separate PR, create another happo story with a gallery view (similar to PR description screenshot) and re-run this job, to try to catch if there are more diffs.
- For the upload/arrow-up, I think we should just check with livegrep if we have any usage of them and fix them as well, afterward. 
- I don't know if we have to keep both close and close-clear, but if we do, I think we need to replace tag deletable state icon with close-clear

esmailpour-hosein (approved):

PR #474, opened 2022-04-29
feat(select): add label and id for select component

esmailpour-hosein:
Add label and id for select component, to have consistent look and properties as text-field component.
Screenshot 2022-04-29 at 09 51 19


tradeshift-larsbojensen (approved):
Great job!

PR #477, opened 2022-05-05
PEAA-1071: migrate to node 16 and add overrides of vulnerable packages

aleks-elkin:
- Migrated to node 16 to use `overrides` functionality from npm v8.
- Regenerated package-lock to version 2.
- Added overrides for these packages:
  -   "highlight.js": ">=10.4.1"
  -  "glob-parent": ">=5.1.2"
  -  "nth-check": ">=2.0.1"
- Removed an override of the package "set-value" because we no longer use vulnerable version.


esmailpour-hosein (approved):
🚀

PR #481, opened 2022-06-15
chore: update lerna to v5

tradeshift-larsbojensen:


tradeshift-larsbojensen:
Version 5.1.2 of lerna does not fix the issue after all.

PR #487, opened 2022-06-21
PEAA-1120: add support of inline svg in icon component and use inline svg in other components

aleks-elkin:
- added support of inline svg string in `ts-icon` component
- switched all components that use `ts-icon` internally to use inline svg strings instead of icon's dynamic loading


packages/components/button/src/button.js
```diff
@@ -26,6 +26,8 @@ export class TSButton extends TSElement {
 			busy: { type: Boolean, reflect: true },
 			/** Icon name, see the list of available icons in ts-icon component. Also note that it will hide the slot content unless the type is text */
 			icon: { type: String, reflect: true },
+			/** Icon source path. Also note that it will hide the slot content unless the type is text */
```
esmailpour-hosein, line 29: Can we explain a bit about what's the difference between `iconSrc` and `icon` properties here, and when to use which?

esmailpour-hosein (approved):

PR #488, opened 2022-06-28
PEAA-1117: fix the new release

aleks-elkin:
What was the issue with the new release:
after switch to npm v8 and package-lock v2 (introduced in npm v7) lerna no longer updates the root package-lock file. Apparently it was never supported but for some reason it worked before.

What was done in this PR:
- Removed dependencies section from the global package.json because lerna do not support this setup. I don't know why it worked before, but now with new package-lock format introduced in npm v7 it stop working.
- After removal of global list of dependencies I had to solve the issue that storybook no longer can resolve the imports of the packages that we create a story for. Eventually I solved this with import from `../lib/<package-name>.esm`.
- Because we now don't have all packages in global package-lock we need add dependencies from the stories as well as from src files. E.g. if we use a `button` component only in story we still need to add it to a package.json of this component  so lerna can symlink it to a local package and storybook will successfully resolve it from local folder.
- I had to add lerna bootstrap to several scripts related to build storybook to be sure that all packages are symlinked before we run storybook.
- Also I added a `--hoist` cli key to lerna bootstrap command so it will add all external packages (e.g. lit-element) to a root package-lock instead of generating local package-lock files in child folders.
- added a webpack override in storybook config to correctly resolve imports of icons.


tradeshift-larsbojensen (approved):
This is great stuff! 👍

PR #490, opened 2022-06-30
Fix storybook deploy

aleks-elkin:
After we introduced npm overrides storybook has stop building successfully due to some inconsistency in dependencies.
I added the override for `react-syntax-highlighter` for no longer supported storybook addon `@storybook/addon-notes@5.3.21` to align with other storybook plugins and it seems to solve the issue.
Also, added an override for `trim-newlines` package to address the dependabot alert.

### DIFF
--- a/.storybook-happo/preview-head.html
+++ b/.storybook-happo/preview-head.html
@@ -31,5 +31,5 @@
 	})();
 </script>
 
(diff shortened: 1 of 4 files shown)


tradeshift-larsbojensen (approved):
👍

PR #491, opened 2022-07-04
PEAA-1120: add a check for inlined svg strings when dynamically resolve a module

aleks-elkin:
In some configurations of webpack it can inline svg to dynamically imported modules.
I added a check for this case: if we got valid svg string we will use it as an icon svg, otherwise we will try to use imported string as Url to fetch svg.


tradeshift-larsbojensen (approved):
👍

PR #497, opened 2022-07-22
fix: [VULN-4068] - [elements] terser has vulnerabilities

adinaiordache2017:
fix: [VULN-4068] - [elements] terser has vulnerabilities

[VULN-4068]: https://tradeshift.atlassian.net/browse/VULN-4068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,6 +141,9 @@
     "@storybook/storybook-deployer": {
       "parse-url": "6.0.1"
     },
+    "terser-webpack-plugin": {
+      "terser": "^5.14.2"
+    },
     "trim-newlines": "^3.0.1",
     "glob-parent": ">=5.1.2",
     "nth-check": ">=2.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -144,7 +144,8 @@
     "trim-newlines": "^3.0.1",
     "glob-parent": ">=5.1.2",
     "nth-check": ">=2.0.1",
-    "trim": ">=0.0.3"
+    "trim": ">=0.0.3",
+    "terser": "^5.14.2"
```
esmailpour-hosein, line 148: nit, it would be good idea to add the specific dependency of ours that is depending on `terser` as wrapper, so we know when to get rid of it easier.

esmailpour-hosein (approved):

package.json
adinaiordache2017: I will try to do that

PR #500, opened 2022-08-11
fix: make html attrs to be string or boolean. Support "class" attr instead of "className"

aleks-elkin:
Made two fixes:
- Transform attr's types to be only string or boolean. According to HTML specification, element's attribute (aka `content attribute`) can be `string` or `boolean`: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#content_versus_idl_attributes
- Added a `class` attribute to all elements. In React developers should use `class` instead of `className` for Web Components to assign css classes: https://reactjs.org/docs/web-components.html#using-web-components-in-react


adinaiordache2017 (approved):
👍

PR #504, opened 2022-09-21
PEAA-1317: fix ts-card story and stories order

aleks-elkin:
- fixed error with custom HTML in `ts-card` story. Made it the same way as we have in `ts-board` story.
- Made order of stories alphabetical. Documentation: https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy#sorting-stories


adinaiordache2017 (approved):
👍

PR #505, opened 2022-10-18
types(datepicker): fix datepicker types

seriouslag:
'function' is not a valid type in typescript

### DIFF
--- a/scripts/generate-types.js
+++ b/scripts/generate-types.js
@@ -10,6 +10,20 @@ const {
 	camelize
 } = require('./helpers');
 
+/**
+ * @typedef {Object} JsProperties
+ * @property {string} Property - Indicates whether the Courage component is present.
+ * @property {string} Type - Indicates whether the Power component is present.
+ * @property {string} [Description] - Indicates whether the Wisdom component is present.
+ */
+
+/**
+ *
+ * @param {string} name - name of property
+ * @param {string} type - type of property
+ * @param {string} [description] - description of property
+ * @returns
+ */
 const generatePropertyLine = (name, type, description) => {
 	let result = '';
 	if (description) {
@@ -19,12 +33,7 @@ const generatePropertyLine = (name, type, description) => {
 		// if name has dashes we put it in quotes
 		name = `"${name}"`;
 	}
-	// Replace 'Array' with '[]'.
-	// Better to use Array<T> but we don't have information about structure of array items now.
-	if (type.toLowerCase() === 'array') {
-		type = 'any[]';
-	}
-	result += `\t${name}?: ${type.toLowerCase()};${EOL}${EOL}`;
+	result += `\t${name}?: ${type};${EOL}${EOL}`;
 	return result;
 };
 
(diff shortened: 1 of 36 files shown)


yeetrium (approved):
🌽

aleks-elkin:
Thank you for contribution!
Unfortunately, this file is autogenerated, so you need to fix it in another place. To be precise, you should change the function `generatePropertyLine` here: https://github.com/Tradeshift/elements/blob/master/scripts/generate-types.js#L13

seriouslag:
> Thank you for contribution! Unfortunately, this file is autogenerated, so you need to fix it in another place. To be precise, you should change the function `generatePropertyLine` here: https://github.com/Tradeshift/elements/blob/master/scripts/generate-types.js#L13

Updated generator script and updated all type files.

yeetrium (approved):
⚡

scripts/generate-types.js
```diff
@@ -42,24 +51,67 @@ const getHtmlAttributeInterfaces = (typesFileContent, className, properties) =>
 	typesFileContent += `\tclass?: string;${EOL}`;
 	properties.forEach(property => {
 		// HTML attributes can be one of two types: boolean or string
-		const htmlType = property.Type.toLowerCase() === 'boolean' ? property.Type : 'string';
-		typesFileContent += generatePropertyLine(property.Attribute, htmlType, property.Description);
+		// const htmlType = property.Type.toLowerCase() === 'boolean' ? property.Type : 'string';
+		const tsType = jsTypeToTypescriptType(property.Type);
+		typesFileContent += generatePropertyLine(property.Attribute, tsType, property.Description);
```
aleks-elkin, line 56: Please, revert these changes. According to HTML spec, the accepted types for html attributes are string or boolean.
Actually, you can test yourself with a date-picker's function `isDisabledDate`, this code will not trigger the function:
```ts
  const func = () => {
    console.log('called!');
  }

<ts-date-picker isDisabledDate={func}></ts-date-picker>
```

I think the correct way is to do it via ref:
```ts
import React, {useEffect, useRef} from "react";
import "@tradeshift/elements.date-picker";
import {TSDatePicker} from "@tradeshift/elements.date-picker";

function DatePicker() {
  const dateRef = useRef<TSDatePicker & HTMLElement>(null);
  const func = () => {
    console.log('called!');
  }
  useEffect(() => {
    if (dateRef && dateRef.current) {
      dateRef.current.isDisabledDate = func;
    }
  }, [dateRef]);
  return <ts-date-picker ref={dateRef}></ts-date-picker>;
}
```

aleks-elkin (changes requested):

scripts/generate-types.js
seriouslag: Good catch!

Updated.
I changed 'boolean' to be of a similar type as React does for HTML attributes

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L54

aleks-elkin (approved):
LGTM!

